PR #211, opened 2021-04-15
Fix retrying when connecting to k8s cluster

kevjin:
If all the retries to connect to the k8s cluster fails (ex. rbac restrictions), we shouldn't just proceed onwards with the rest of the code. Raise an exception if all the retries here fail.


NitinAgg (commented):
@juandiegopalomino should we merge this?

PR #253, opened 2021-05-25
Switch to ruamel

juandiegopalomino:
Because I think it may support account ids starting with zero. Will test in a bit. OK it did with a bit of work


opta/core/validator.py
```diff
@@ -78,25 +78,35 @@ def validate(self, value: Any) -> List[str]:
...
+        return _get_yamale_errors(value, schema_path, self.extra_validators)
+
+
+class AwsId(Validator):
+    tag = "aws_id"
+
+    def _is_valid(self, value: Any) -> bool:
+        str_value = str(value)
+
+        return str_value.isdigit() and len(str_value) == 12
```
legalosLOTR, line 90: Let's remove the 12 length constraint? No need to add unnecessary validation.

opta/layer.py
```diff
@@ -140,6 +139,10 @@ def load_from_dict(cls, conf: Dict[Any, Any], env: Optional[str]) -> Layer:
         providers = conf.pop("providers", {})
         if "aws" in providers:
             providers["aws"]["version"] = AWS_PROVIDER_VERSION
+            providers["aws"]["account_id"] = providers["aws"].get("account_id", "")
+            account_id = str(providers["aws"]["account_id"])
+            account_id = "0" * (12 - len(account_id)) + account_id
```
legalosLOTR, line 144: uh what? If it's a string now - why would the 0 get chopped off?

legalosLOTR (commented):
Actually, why do we need ruamel? Can't we just make aws_id be a string?

opta/core/validator.py
juandiegopalomino: AWS defines account ids as having EXACTLY 12 digits

legalosLOTR: That can change down the line... or there can be weird exceptions where this doesn't apply. No big advantage of having this constraint...

opta/layer.py
juandiegopalomino: Good question, here's what is happening:
The old yaml library was using yaml v 1.1 where any integer starting with 0 and the digits are less than 8 is transformed to octal https://stackoverflow.com/questions/54820256/how-to-read-load-yaml-parameters-with-leading-zeros-as-a-string
Ruamel uses the v1.2 yaml spec and maintains the base-10 value of the number with leading zeroes (thereby not scrambling them) but still converts it as an int w/ no leading zeros (we can keep the leading zeros by using the base loader in which EVERY value is kept as a string, which we don't want). With that integer we can then transform it to a string with leading zeros by adding zeros until the length is 12 as all aws account ids have exactly 12 digits.

legalosLOTR: Is it possible to move this logic to somewhere "lower" in the stack? I'm just wondering if we end up using aws_id somewhere else - we will forget to add this padding.

Maybe in utils.py where you are importing ruamel?

juandiegopalomino: I thought about that-- this is as low as we can go. We always create layers from loading from yaml or dict (and loading from yaml immediately calls loading from dict). That way the id is already fixed _before_ the layer class is even made.

opta/core/validator.py
juandiegopalomino: done

legalosLOTR (approved):
Great. Let’s make sure to thoroughly test this before releasing! @NitinAgg

PR #254, opened 2021-05-25
Fixing check for upgrade

juandiegopalomino:


legalosLOTR (commented):
Can you also move the semver check (on line 44) inside the try-catch? So we don't have future problems like this.

legalosLOTR (approved):

PR #255, opened 2021-05-26
Redoing image catch

juandiegopalomino:


opta/commands/apply.py
```diff
@@ -130,6 +130,42 @@ def _apply(
...
+                    if auto_approve
+                    else click.confirm(
+                        f"WARNING There is an existing deployment (tag={current_tag}) and the pods will be killed as you "
+                        f"did not specify an image tag. Would you like to keep the existing deployment alive? (y/n)",
+                    )
+                )
+                if response:
+                    image_tag = current_tag
+                else:
+                    raise RuntimeError("Aborting")
```
legalosLOTR, line 165: In the else case, we need to continue the deployment and kill the pods.

opta/commands/apply.py
```diff
@@ -130,6 +130,42 @@ def _apply(
...
+            current_tag = current_image_tag(layer)
+            if (
+                current_tag is not None
+                and image_tag is None
+                and service_module.data.get("image", "") == "AUTO"
+                and not test
+            ):
+                response = (
+                    True
+                    if auto_approve
```
legalosLOTR, line 156: I thought we said auto_approve won't apply in this situation

opta/commands/apply.py
```diff
@@ -130,6 +130,42 @@ def _apply(
...
+    service_modules = (
+        layer.get_module_by_type("k8s-service")
+        if layer.cloud == "aws"
+        else layer.get_module_by_type("gcp-k8s-service")
+    )
+
+    if len(service_modules) > 0 and (
+        Terraform.downloaded_state.get(layer.name, False)
+        or Terraform.download_state(layer)
+        and get_cluster_name(layer) is not None
```
legalosLOTR, line 142: When does this happen? And do we need to handle that?

juandiegopalomino: This only happens in the first run with the k8s-service module and given that it can't have preexisting pods (because it is just being deployed now) there's nothing to handle

opta/commands/apply.py
juandiegopalomino: I thought we said it would?

opta/commands/apply.py
juandiegopalomino: done

opta/commands/apply.py
```diff
@@ -185,7 +219,6 @@ def _apply(
                 )
                 and get_cluster_name(layer) is not None
             ):
-                configure_kubectl(layer)
```
legalosLOTR, line 188: Intentional?

opta/commands/apply.py
legalosLOTR: Naah. I don’t think we should allow users to kill their pods with auto approve. That has to be a manual operation.

opta/commands/apply.py
juandiegopalomino: yes

legalosLOTR (approved):

PR #258, opened 2021-06-01
Stable secrets

juandiegopalomino:
Still there after rollback. Different secret resource between linked and manual secrets

### DIFF
--- a/config/tf_modules/gcp-k8s-service/k8s-service/templates/deployment.yaml
+++ b/config/tf_modules/gcp-k8s-service/k8s-service/templates/deployment.yaml
@@ -1,10 +1,10 @@
 {{- if ne ((keys .Values.port) | len) 1 }}
   {{- fail "Port must specify either a single tcp port (for websockets) or a single http port" }}
-{{- end }}
-{{- if not ( or (hasKey .Values.port "tcp") ( or (hasKey .Values.port "http") (hasKey .Values.port "grpc"))) }}
+  {{- end }}
+  {{- if not ( or (hasKey .Values.port "tcp") ( or (hasKey .Values.port "http") (hasKey .Values.port "grpc"))) }}
   {{- fail "Port must specify either a single tcp port (for websockets) or a single http port" }}
-{{- end }}
-{{- if .Values.deployPods }}
+  {{- end }}
+  {{- if .Values.deployPods }}
 apiVersion: apps/v1
 kind: Deployment
 metadata:
@@ -12,12 +12,12 @@ metadata:
   labels:
     tags.datadoghq.com/service: {{ include "k8s-service.serviceName" . }}-{{ include "k8s-service.namespaceName" . }}
     tags.datadoghq.com/version: {{ .Values.version | quote }}
-    {{- include "k8s-service.labels" . | nindent 4 }}
+  {{- include "k8s-service.labels" . | nindent 4 }}
   namespace: {{ include "k8s-service.namespaceName" . }}
 spec:
   selector:
     matchLabels:
-      {{- include "k8s-service.selectorLabels" . | nindent 6 }}
+  {{- include "k8s-service.selectorLabels" . | nindent 6 }}
   template:
     metadata:
       labels:
@@ -42,31 +42,36 @@ spec:
           ports:
             - name: main
               protocol: TCP
-            {{- if hasKey .Values.port "http" }}
+              {{- if hasKey .Values.port "http" }}
               containerPort: {{ .Values.port.http }}
-            {{- end }}
-            {{- if hasKey .Values.port "tcp" }}
+              {{- end }}
+              {{- if hasKey .Values.port "tcp" }}
               containerPort: {{ .Values.port.tcp }}
-            {{- end }}
-            {{- if hasKey .Values.port "grpc" }}
+              {{- end }}
+              {{- if hasKey .Values.port "grpc" }}
               containerPort: {{ .Values.port.grpc }}
-            {{- end }}
+          {{- end }}
           env:
             - name: RDS_CA_PATH # This is the path to the public key for the docdb tls
               value: "/config/rds_ca.pem"
-            {{- if or (.Values.envVars) (.Values.secrets) }}
             {{ range .Values.envVars }}
             - name: {{ .name | quote }}
               value: {{ .value | quote }}
             {{ end }}
-            {{ range $val := .Values.secrets }}
+            {{ range $val := .Values.linkSecrets }}
             - name: {{ $val.name | quote }}
               valueFrom:
                 secretKeyRef:
                   name: secret
                   key: {{ $val.name | quote }}
             {{ end }}
-            {{- end }}
+            {{ range $val := .Values.manualSecrets }}
+            - name: {{ $val | quote }}
+              valueFrom:
+                secretKeyRef:
+                  name: manual-secrets
+                  key: {{ $val | quote }}
+          {{ end }}
           {{/* No probes for grpc at the moment */}}
           {{ if not (hasKey .Values.port "grpc") }}
           livenessProbe:
@@ -79,7 +84,7 @@ spec:
             httpGet:
               port: main
               path: {{ .Values.livenessProbePath }}
-            {{ end }}
+          {{ end }}
           readinessProbe:
             initialDelaySeconds: 30
             periodSeconds: 10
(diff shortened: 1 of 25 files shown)


config/tf_modules/gcp-k8s-service/k8s-service/templates/deployment.yaml
```diff
@@ -42,31 +42,36 @@ spec:
...
+            {{ range $val := .Values.linkSecrets }}
             - name: {{ $val.name | quote }}
               valueFrom:
                 secretKeyRef:
                   name: secret
                   key: {{ $val.name | quote }}
             {{ end }}
-            {{- end }}
+            {{ range $val := .Values.manualSecrets }}
+            - name: {{ $val | quote }}
```
legalosLOTR, line 69: Shouldn't this be `$val.name` like above?

opta/core/kubernetes.py
```diff
@@ -217,6 +222,82 @@ def current_image_tag(layer: "Layer",) -> Optional[str]:
     return None
 
 
+def create_namespace(layer_name: str) -> None:
```
legalosLOTR, line 225: Should be named `create_namespace_if_not_exists`

opta/core/kubernetes.py
```diff
@@ -217,6 +222,82 @@ def current_image_tag(layer: "Layer",) -> Optional[str]:
...
+        v1.create_namespace(
+            body=V1Namespace(
+                metadata=V1ObjectMeta(
+                    name=layer_name, annotations={"linkerd.io/inject": "enabled"}
+                )
+            )
+        )
+
+
+def create_manual_secrets(layer_name: str) -> None:
```
legalosLOTR, line 239: Should be named `create_manual_secrets_if_not_exists`

opta/commands/secret.py
```diff
@@ -84,12 +81,10 @@ def update(secret: str, value: str, env: Optional[str], config: str) -> None:
     _raise_if_no_k8s_cluster_exists(layer)
 
     configure_kubectl(layer)
+    create_namespace(layer.name)
+    create_manual_secrets(layer.name)
```
legalosLOTR, line 85: I think it would be cleaner to call `create_manual_secrets` from `update_manual_secrets` itself

opta/commands/secret.py
```diff
@@ -38,15 +42,14 @@ def view(secret: str, env: Optional[str], config: str) -> None:
     _raise_if_no_k8s_cluster_exists(layer)
 
     configure_kubectl(layer)
-    load_kube_config()
-    v1 = CoreV1Api()
-    api_response = v1.read_namespaced_secret("secret", layer.name)
-    if secret not in api_response.data:
+    create_namespace(layer.name)
+    secrets = deep_merge(get_manual_secrets(layer.name), get_linked_secrets(layer.name))
```
legalosLOTR, line 46: Can we just create a `get_secrets` function in kubernetes.py? Instead of doing this merge logic at every place?

opta/commands/secret.py
```diff
@@ -38,15 +42,14 @@ def view(secret: str, env: Optional[str], config: str) -> None:
...
+    if secret not in secrets:
```
legalosLOTR, line 47: So this can happen in multiple situations:
1) You haven't run an apply yet - so the linked secrets are not available. Solution here is to run apply
2) You haven't set a manual secret using the cli. Solution here is to run `secret update`
3) You don't have this manual secret in the yaml.

I think we should print the appropriate message in each of the cases. Instead of just saying "You sure you set it in your yaml".

opta/module_processors/aws_k8s_service.py
```diff
@@ -85,13 +87,19 @@ def process(self, module_idx: int) -> None:
         }
         if "image_tag" in self.layer.variables:
             self.module.data["tag"] = self.layer.variables["image_tag"]
+        seen = set()
```
legalosLOTR, line 90: What are you trying to do here? Why would we have duplicates in `link_secrets`?

legalosLOTR:
Test cases we need to verify (no need to do in this PR but must be done before releasing @NitinAgg):
- Deploy a service with secrets before this change and try to upgrade it with this change. Make sure the user gets prompted to set the secrets again.
- set secrets before apply, make sure the container gets them on apply
- update secrets after apply, make sure next deploy gets them
- update secrets after apply, introduce deploy failure, do a successful deploy and make sure container gets them

config/tf_modules/gcp-k8s-service/k8s-service/templates/deployment.yaml
juandiegopalomino: No in manual secrets we only pass the list of secret names because there is no need for the helm chart to know the values.

opta/module_processors/aws_k8s_service.py
juandiegopalomino: It's a long story-- basically the process function get called multiple times with the same stateful module object and so this deduplicates secrets being set from the different calls

opta/commands/secret.py
juandiegopalomino: done

opta/commands/secret.py
juandiegopalomino: done

opta/core/kubernetes.py
juandiegopalomino: done

opta/core/kubernetes.py
juandiegopalomino: done

opta/commands/secret.py
juandiegopalomino: `You haven't run an apply yet - so the linked secrets are not available. Solution here is to run apply` -- I'm afraid that there's no way of telling what is a valid linked secret and what is not. Sure, we can try and parse the yaml again and see all the manual secrets, but we won't know if a given string is a valid linked secret or just baloney.

Going over whether this is a manual secret or not is also a tad tedious-- I can reform this to give an informative error message such as "You specified to see secret X but currently we only have secrets [Y ,Z]-- please use the opta secret update command to add the new secrets!"

legalosLOTR: Ah. Let's do this instead: `We couldn't find a secret named <xyz>. You either need to add it to your opta.yml file or if it's already there - update it via secret update.`

opta/module_processors/aws_k8s_service.py
legalosLOTR: Ouch that sounds like it will bite us down the line. We shouldn't maintain stateful objects.
But not a blocker for this PR!

opta/core/kubernetes.py
```diff
@@ -222,12 +222,12 @@ def current_image_tag(layer: "Layer",) -> Optional[str]:
     return None
 
 
-def create_namespace(layer_name: str) -> None:
+def create_namespace_if_not_exists(layer_name: str) -> None:
     load_kube_config()
     v1 = CoreV1Api()
     namespaces = v1.list_namespace(field_selector=f"metadata.name={layer_name}")
     if len(namespaces.items) == 0:
```
legalosLOTR, line 229: Do we need this check if we're using `v1.create_namespace_if_not_exists`?

opta/core/kubernetes.py
```diff
@@ -236,14 +236,14 @@ def create_namespace(layer_name: str) -> None:
         )
 
 
-def create_manual_secrets(layer_name: str) -> None:
+def create_manual_secrets_if_not_exists(layer_name: str) -> None:
     load_kube_config()
     v1 = CoreV1Api()
     manual_secrets: V1SecretList = v1.list_namespaced_secret(
```
legalosLOTR, line 242: Wouldn't this error if namespace doesn't exist?

opta/module_processors/datadog.py
```diff
@@ -38,7 +38,7 @@ def process(self, module_idx: int) -> None:
             field_selector=f"metadata.name={self.layer.name}"
         )
         if len(namespaces.items) == 0:
-            self.v1.create_namespace(
+            self.v1.create_namespace_if_not_exists(
```
legalosLOTR, line 41: Is it possible to use the methods from `kubernetes.py` here instead of calling k8s directly?

opta/core/kubernetes.py
juandiegopalomino: yes? K8s returns a 409 error otherwise and this seemed cleaner than handling the 409

opta/module_processors/datadog.py
juandiegopalomino: oh crap my bad-- did not mean to update this one

opta/core/kubernetes.py
legalosLOTR: ah wondering if the 409 will lead to an exception on line 228?

opta/core/kubernetes.py
juandiegopalomino: which is why we call create_namespace_if_not_exists in secrets.py prior to the update_manual_secrets call

legalosLOTR (approved):
lgtm - once you update the error message

PR #261, opened 2021-06-02
Clean state

juandiegopalomino:
No reused .terraform.tfstate
Remote states are cleaned up now on destroy

### DIFF
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -3,6 +3,7 @@ from typing import TYPE_CHECKING, List, Optional, TypedDict
 
 import boto3
 from botocore.config import Config
+from mypy_boto3_dynamodb import DynamoDBClient
 
 from opta.utils import fmt_msg, logger
 
@@ -77,6 +78,23 @@ class AWS:
         if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
             raise Exception(f"Failed to delete opta config in {bucket}/{config_path}.")
 
+        s3_client = boto3.client("s3")
+        resp = s3_client.delete_object(Bucket=bucket, Key=self.layer.name)
+        if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
+            raise Exception(
+                f"Failed to delete opta config in {bucket}/{self.layer.name}."
+            )
+
+        providers = self.layer.gen_providers(0)
+        dynamodb_table = providers["terraform"]["backend"]["s3"]["dynamodb_table"]
+
+        dynamodb_client: DynamoDBClient = boto3.client(
+            "dynamodb", config=Config(region_name=self.region)
+        )
+        dynamodb_client.delete_item(
+            TableName=dynamodb_table,
+            Key={"LockID": {"S": f"{bucket}/{self.layer.name}-md5"}},
+        )
         logger.info("Deleted opta config from s3")
 
     @staticmethod
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -78,13 +78,10 @@ class AWS:
         if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
             raise Exception(f"Failed to delete opta config in {bucket}/{config_path}.")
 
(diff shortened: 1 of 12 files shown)


opta/core/aws.py
```diff
@@ -77,6 +78,23 @@ def delete_opta_config(self) -> None:
         if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
             raise Exception(f"Failed to delete opta config in {bucket}/{config_path}.")
 
+        s3_client = boto3.client("s3")
+        resp = s3_client.delete_object(Bucket=bucket, Key=self.layer.name)
+        if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
```
legalosLOTR, line 83: Add a comment here to explain what this if implies?

opta/core/aws.py
```diff
@@ -77,6 +78,23 @@ def delete_opta_config(self) -> None:
...
+                f"Failed to delete opta config in {bucket}/{self.layer.name}."
+            )
+
+        providers = self.layer.gen_providers(0)
+        dynamodb_table = providers["terraform"]["backend"]["s3"]["dynamodb_table"]
+
+        dynamodb_client: DynamoDBClient = boto3.client(
+            "dynamodb", config=Config(region_name=self.region)
+        )
+        dynamodb_client.delete_item(
```
legalosLOTR, line 94: Why not just delete the whole dynamodb table?

opta/core/aws.py
```diff
@@ -77,6 +78,23 @@ def delete_opta_config(self) -> None:
         if resp["ResponseMetadata"]["HTTPStatusCode"] != 204:
             raise Exception(f"Failed to delete opta config in {bucket}/{config_path}.")
 
+        s3_client = boto3.client("s3")
```
legalosLOTR, line 81: (if i'm understanding the code structure properly) Deleting the s3 and dynamodb should be in `destroy_all()` and not in `delete_opta_config`

opta/core/kubernetes.py
```diff
@@ -165,10 +165,10 @@ def _aws_configure_kubectl(layer: "Layer") -> None:
             )
         )
 
-    cluster_name = get_cluster_name(layer)
+    cluster_name = get_cluster_name(root_layer)
 
     if cluster_name is None:
-        raise Exception("The EKS cluster name could not be determined.")
+        raise UserErrors("The EKS cluster name could not be determined.")
```
legalosLOTR, line 171: Should this be a user error?

opta/core/aws.py
juandiegopalomino: Because the table has the locks for all other services in the environment and the environment as well. We only care about removing the lock for the current service or the environment (and if we're doing it for the environment then that means we would have already cleared the locks for the services and the next step would be deleting the fully empty table-- totally graceful)

opta/core/aws.py
juandiegopalomino: I see where you were going, but actually, it should be on it's own function (where I just moved it)

opta/core/aws.py
juandiegopalomino: Check out new function

opta/core/kubernetes.py
juandiegopalomino: my bad

opta/core/terraform.py
```diff
@@ -351,8 +347,13 @@ def download_state(cls, layer: "Layer") -> bool:
         else:
             raise UserErrors("Need to get state from S3 or GCS")
 
-        cls.downloaded_state[layer.name] = True
-        return True
+        with open(state_file, "r") as file:
+            raw_state = file.read().replace("\n", "")
```
legalosLOTR, line 351: why do we need this? yaml is whitespace sensitive - so won't it get messed up by this?

juandiegopalomino: `if raw_state != ""`

legalosLOTR (approved):

PR #267, opened 2021-06-03
Added support for mysql on gcp

juandiegopalomino:
Screen Shot 2021-06-03 at 4 15 40 PM
Tested with default engine 5.7

### DIFF
--- a/config/tf_modules/gcp-mysql/main.tf
+++ b/config/tf_modules/gcp-mysql/main.tf
@@ -7,7 +7,9 @@ resource "random_password" "root_auth" {
   special = false
 }
 
-# TODO: add encryption key name once out of beta
+# TODO: currently we are not performing disk encryption as the feature is currently in beta. You may read more about it
+# here: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/sql_database_instance#encryption_key_name
+# We currently do not wish to tangle with the google-beta provider so we are skipping this feature for now.
 resource "google_sql_database_instance" "instance" {
   name                = "opta-${var.layer_name}-${var.module_name}-${random_id.key_suffix.hex}"
   database_version    = "MYSQL_${var.engine_version}"
--- a/config/tf_modules/gcp-mysql/main.tf
+++ b/config/tf_modules/gcp-mysql/main.tf
@@ -1,0 +1,49 @@
+resource "random_id" "key_suffix" {
+  byte_length = 8
+}
+
+resource "random_password" "root_auth" {
+  length  = 20
+  special = false
+}
+
+# TODO: add encryption key name once out of beta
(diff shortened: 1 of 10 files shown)


config/tf_modules/gcp-mysql/main.tf
```diff
@@ -0,0 +1,49 @@
+resource "random_id" "key_suffix" {
+  byte_length = 8
+}
+
+resource "random_password" "root_auth" {
+  length  = 20
+  special = false
+}
+
+# TODO: add encryption key name once out of beta
```
legalosLOTR, line 10: what does this mean?

legalosLOTR (approved):
Can you also send out a pr for the relevant docs? :))

config/tf_modules/gcp-mysql/main.tf
juandiegopalomino: Disk encryption in GCP is currently in beta

config/tf_modules/gcp-mysql/outputs.tf
```diff
@@ -0,0 +1,16 @@
+output "db_user" {
```
legalosLOTR, line 1: Add a comment here that this list of outputs needs to be the same as the ones for gcp-postgres and should match the ones at `gcp_k8s_service.py:handle_rds_link()`

config/tf_modules/gcp-mysql/main.tf
juandiegopalomino: So I decided not to tangle with that and the _separate_ provider it would make us use

legalosLOTR: Ah so is it currently encrypted or not?

legalosLOTR: Please clarify in the comment that "disk encryption in gcp mysql is currently in beta"

config/tf_modules/gcp-mysql/outputs.tf
juandiegopalomino: done!

config/tf_modules/k8s-service/k8s-service/templates/deployment.yaml
```diff
@@ -25,7 +25,7 @@ spec:
         tags.datadoghq.com/service: {{ include "k8s-service.serviceName" . }}-{{ include "k8s-service.namespaceName" . }}
         tags.datadoghq.com/version: {{ .Values.version | quote }}
       annotations:
-        config.linkerd.io/skip-outbound-ports: "8126,5432"
+        config.linkerd.io/skip-outbound-ports: "8126,5432,3306"
```
NitinAgg, line 28: Let's add a comment for each of the ports, why we had to skip them

PR #269, opened 2021-06-04
Implement CORS support in the s3 module

legalosLOTR:
# Description
Pretty straightforward
TODO: update docs

# Test plan
Try creating a s3 bucket via opta, with various permutations of the cors_rule and make sure it works


config/tf_modules/aws-s3/variables.tf
```diff
@@ -27,3 +27,7 @@ variable "bucket_policy" {
   default = null
 }
 
+variable "cors_rule" {
+  type = any
+  default = null
```
juandiegopalomino, line 32: why not default = []?

config/schema/modules/aws-s3.yaml
```diff
@@ -3,6 +3,12 @@ name: str(required=False)
 env_name: str(required=False)
 layer_name: str(required=False)
 module_name: str(required=False)
-bucket_name: str(required=False)
+bucket_name: str(required=True)
 block_public: any(required=False)
-bucket_policy: any(required=False)
\ No newline at end of file
+bucket_policy: any(required=False)
+cors_rule:
```
juandiegopalomino, line 9: is this required=False by default?

config/tf_modules/aws-s3/variables.tf
legalosLOTR: It's a map, not a list

config/schema/modules/aws-s3.yaml
legalosLOTR: Ah it's not! Let me figure out a way to fix it

config/tf_modules/aws-s3/bucket.tf
```diff
@@ -8,6 +8,17 @@ resource "aws_s3_bucket" "bucket" {
     }
   }
   force_destroy = true
+
+  dynamic "cors_rule" {
+    for_each = [var.cors_rule]
```
juandiegopalomino, line 13: for_each = var.cors_rule == null ? [] : [var.cors_rule]

juandiegopalomino (approved):

PR #270, opened 2021-06-07
Fixing tls

juandiegopalomino:
OK so this is why we were having issues with public-uri false.

Cert manager gives a k8s user the ability to issue ssl certificates based on an issues "[type](https://cert-manager.io/docs/configuration/)" (e.g. ACME, self-signed, custom made certificate authority, vault, etc..). There's a one-time step of setting up the issuer and from there your ingresses can ask (via an annotation) "hey I want a certificate from this issuer I created with cert manager" -- cert manager then reads the ingress spec and figures out what hosts the cert is for and says "ok I'll create a fresh new one and put it where you said you will look".

The problem that we were encountering is that if the public-uri is all (e.g. support all hosts) then we tried to make the ssl cert for a wild card "*" which is not supported. Given that the ssl cert is just used to make our communications between the nginx ingress pods and the cloud load balancer encrypted (needed by grpc but also good security practices) and the load balancers do not care about domain verification, the host is totally inconsequential. The fix is to fallback on some determinist dns name and for consistency we now will use the service-name.namespace as that's how the service is identified within the k8s cluster.

One future change we are preparing for: currently, the certificates used are fully self-signed and do not belong to any CA. This _feels_ unclean and to remedy this we are setting up a custom-ca issuer with certmanager so that in the future all certs can be managed via this one CA. We are not switching services to use it right now as that would make service deployments fail until the environment is updated and new issuer created

### DIFF
--- a/config/tf_modules/gcp-k8s-base/opta-base/Chart.yaml
+++ b/config/tf_modules/gcp-k8s-base/opta-base/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.2.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
--- a/config/tf_modules/gcp-k8s-base/opta_base.tf
+++ b/config/tf_modules/gcp-k8s-base/opta_base.tf
@@ -2,11 +2,6 @@ resource "helm_release" "opta_base" {
   chart     = "${path.module}/opta-base"
   name      = "opta-base"
   namespace = "default"
-  values = [
-    yamlencode({
-      version : 1
-    })
-  ]
   depends_on = [
     helm_release.cert_manager
   ]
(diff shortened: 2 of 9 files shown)


config/tf_modules/gcp-k8s-base/opta_base.tf
```diff
@@ -2,6 +2,11 @@ resource "helm_release" "opta_base" {
   chart     = "${path.module}/opta-base"
```
legalosLOTR, line 2: Actually, doesn't helm have a concept of versions?

juandiegopalomino: Yes, but it doesn't really work with terraform:
https://github.com/hashicorp/terraform-provider-helm/issues/102

I tried it myself and you'd need to update the version in the tf template too. This way you only need to update in one place, but I understand if you want to do it the other way.

config/tf_modules/gcp-k8s-base/opta-base/Chart.yaml
```diff
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 0.2.0
```
legalosLOTR, line 18: wooh! this is perfect!

legalosLOTR (approved):

PR #271, opened 2021-06-07
add opta version tag to all resources

juandiegopalomino:
that was easy lol


legalosLOTR:
What's the goal here? This will make every resource get updated - so we should only do this if it's absolutely necessary.

juandiegopalomino:
@legalosLOTR https://run-x.atlassian.net/browse/RUNX-490

PR #274, opened 2021-06-10
Reorganize examples and add airflow example

legalosLOTR:


juandiegopalomino (commented):
1. could we also have aws-example subdomain so that it looks normalized?
2. We're deleting/removing a few envs here pls run opta destroy so that we don't keep paying for them :P

juandiegopalomino:
@legalosLOTR python test failed because 2 tests were using the example yaml files just refactored. Should be easy fix.

legalosLOTR:
> could we also have aws-example subdomain so that it looks normalized?

Not sure I get what you're saying. Can you elaborate?

> We're deleting/removing a few envs here pls run opta destroy so that we don't keep paying for them :P

No environments have been deleted. Just moved around. (i just verified)

juandiegopalomino:
The current domains are example.runx.dev and gcp-example.runx.dev . Might look nicer if it was aws-example.runx.dev and gcp-example.runx.dev.

juandiegopalomino:
@juandiegopalomino it looks like the hybrid environments where destroyed

legalosLOTR:
> The current domains are example.runx.dev and gcp-example.runx.dev . Might look nicer if it was aws-example.runx.dev and gcp-example.runx.dev.

I'm gonna leave that as is - given that dns change is pretty destructive.

> @juandiegopalomino it looks like the hybrid environments where destroyed

Ah you are right! deleting now.

juandiegopalomino (approved):

PR #280, opened 2021-06-14
Allow user to rename db/cache cred envars in the link

juandiegopalomino:
Did it for redis, postgres, and docdb in AWS and redis, postgres, and mysql in GCP

### DIFF
--- a/opta/module_processors/aws_k8s_service.py
+++ b/opta/module_processors/aws_k8s_service.py
@@ -1,4 +1,4 @@
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, Any, List
 
 from opta.core.kubernetes import create_namespace_if_not_exists, get_manual_secrets
 from opta.exceptions import UserErrors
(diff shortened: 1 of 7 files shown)


opta/module_processors/aws_k8s_service.py
```diff
@@ -96,12 +96,25 @@ def process(self, module_idx: int) -> None:
         super(AwsK8sServiceProcessor, self).process(module_idx)
 
     def handle_rds_link(
-        self, linked_module: "Module", link_permissions: List[str]
+        self, linked_module: "Module", link_permissions: List[Any]
     ) -> None:
-        for key in ["db_user", "db_name", "db_password", "db_host"]:
+        required_db_vars = ["db_user", "db_name", "db_password", "db_host"]
+        renamed_vars = {}
+        if len(link_permissions) > 0:
```
legalosLOTR, line 103: Can we pull this out in a common function? instead of copy pasting everywhere? A lot of this linking code could be pulled out I think.

legalosLOTR (commented):
Great job!

opta/module_processors/aws_k8s_service.py
juandiegopalomino: I was thinking about this -- my hesitation was in keeping informative error messages custom to each resource

legalosLOTR: The error messages seem pretty similar and some loss in customization there seems worth it - for code clarity. Just my 2c. Totally fine to skip for now!

legalosLOTR (approved):

PR #281, opened 2021-06-15
Shouldn't have to write name/value in env_vars

juandiegopalomino:


legalosLOTR (approved):
Awesome! Can you file tickets to update the docs for this pr and the last one?

juandiegopalomino:
@legalosLOTR yup, filed new ticket!

PR #282, opened 2021-06-15
Added support for spot instances via new bool

juandiegopalomino:
Can be specified in the eks cluster setup for the default nodegroup or for extra nodegroup. Switch to and from spot is seamless thanks to new hex suffix and create before destroy lifecycle hook. This also fixes the bug of multi-step node ec2 instance type change!
Tested it and encountered no problem.


config/tf_modules/aws-eks/default_node_group.tf
```diff
@@ -35,11 +35,22 @@ resource "aws_iam_role_policy_attachment" "node_group_AmazonEC2ContainerRegistry
...
+  }
+}
+
 resource "aws_eks_node_group" "node_group" {
   cluster_name    = aws_eks_cluster.cluster.name
-  node_group_name = "opta-${var.layer_name}-default"
+  node_group_name = "opta-${var.layer_name}-default-${random_id.key_suffix.hex}"
   node_role_arn   = aws_iam_role.node_group.arn
   subnet_ids      = aws_eks_cluster.cluster.vpc_config[0].subnet_ids
+  capacity_type   = var.spot_instances ? "SPOT" : "ON_DEMAND"
```
legalosLOTR, line 52: Don't we also need to specify the spot price somewhere?

legalosLOTR (approved):
Nice!

config/tf_modules/aws-eks/default_node_group.tf
juandiegopalomino: surprisingly, no. I've heard that were important changes to spot instances and tell you what I'll read up about it right now and report

juandiegopalomino: OK I got some answers:
So AWS released support for spot instances for managed node groups on Dec 1, 2020 (~6 months ago) where the only option they added was this "capacity_type" toggle. Their backend handles everything else and sets the maximum spot price equal to the on demand price (so the ec2s will all terminate when the price goes over). I then noticed _another_ interesting feature: an ASG can be a mix of spot and on demand in some set proportion or backup base. This feature seems supremely useful/obvious and am glad that there is this capability. Can we use it in managed node groups? Sadly, no, not at the moment. But(!) there is an AWS roadmap issue for it: https://github.com/aws/containers-roadmap/issues/1297 . Please like it and ask everyone to like this. Hopefully we'll get it soon.

Screen Shot 2021-06-15 at 11 33 40 AM

PR #286, opened 2021-06-18
Updating high availability settings

juandiegopalomino:
* high_availability setting is split for linkerd_high_availability and nginx_high_availability
* High availability configs are false by default
* Added linkerd_enabled variable to k8s-base with which to disable linkerd (linkerd is enabled by default)


legalosLOTR (approved):
make sure to file a ticket for updating docs

PR #287, opened 2021-06-21
Simplify airflow example

legalosLOTR:
- Remove flower (which is a monitoring tool)
- Enable example dags (this way you can try something from the UI)
- Show worker replicas option


juandiegopalomino (approved):
Why no flower out of curiosity ? Might be fun to show off

legalosLOTR:
2 reasons: 1) It didn't work on the first try so would probably need much more debugging 2) I wanna keep the opta.yml as simple as possible

PR #297, opened 2021-07-08
Add validation for org name

legalosLOTR:
# Description
- Jira: https://run-x.atlassian.net/browse/RUNX-549
- Picked a constraint that works for all 3 aws, gcp and azure. Also made sure we don't have existing org names that violate this (using amplitude)

# Test plan
- Try it out with some valid and invalid org names


juandiegopalomino (commented):
CN not start with a dash

legalosLOTR:
> CN not start with a dash

What do you mean? Starting with a dash seems to work in both aws and gcp. (at least for the bucket creation)

NitinAgg (approved):

juandiegopalomino (approved):

PR #303, opened 2021-07-19
Userdata support for eks nodegroup

juandiegopalomino:


config/tf_modules/aws-eks/default_node_group.tf
```diff
@@ -41,9 +41,28 @@ resource "random_id" "key_suffix" {
...
+  count         = (length(var.node_launch_template) > 0) ? 1 : 0
+  instance_type = var.node_instance_type
+  name_prefix   = "opta-${var.layer_name}"
+  # https://github.com/hashicorp/terraform-provider-aws/issues/15118
+
+  block_device_mappings {
+    device_name = "/dev/sda1"
+
+    ebs {
+      volume_size = var.node_disk_size + 10
```
legalosLOTR, line 58: Why +10?

config/tf_modules/aws-eks/variables.tf
```diff
@@ -1,5 +1,20 @@
...
 
+locals {
+  user_data_prefix = <<EOT
+MIME-Version: 1.0
+Content-Type: multipart/mixed; boundary="==MYBOUNDARY=="
+
+--==MYBOUNDARY==
+Content-Type: text/x-shellscript; charset="us-ascii"
+
+#!/bin/bash
```
legalosLOTR, line 12: I feel like this should be specified by the user

config/tf_modules/aws-eks/default_node_group.tf
juandiegopalomino: A realization that disk is _really_ cheap and that extra space will be needed for the system. Can revert if you don't like

config/tf_modules/aws-eks/variables.tf
juandiegopalomino: Nope, it MUST always be there or else it fails with an ugly error message

juandiegopalomino: Best just to have it there from the beginning as most users won't be clever enough to figure out that they need this-- support the golden path

config/tf_modules/aws-eks/default_node_group.tf
legalosLOTR: I think we should revert. If someone explicitly asks for 10G they should get exactly that.

config/tf_modules/aws-eks/variables.tf
legalosLOTR: It’s a rare enough use case that customizability matters here more than golden path. So id say remove all of this and we can ask the user to specify the multi part stuff and bin/bash in the docs - that’s what aws docs do as well.

legalosLOTR (approved):

PR #306, opened 2021-07-22
Azure SSL via BYO Cert module

juandiegopalomino:
In order to support SSL in Azure we decided to introduce a new module type called byo-cert. With it, a user will be able to add a certificate from their local machine into their opta system for use within the ingress setup. Usage of the module looks like this:

```yaml
name: azure-live-example
org_name: runx
providers:
  azurerm:
    location: centralus
    tenant_id: e1cebc17-5d81-4d16-b435-702e291d395b
    subscription_id: 0fca99ae-5717-4a86-a891-6d64d2a04788
modules:
  - type: base
  - type: byo-cert
    domain: "baloney.runx.dev"
    private_key_file: "./privkey.pem"
    certificate_body_file: "./cert_body.pem"
    certificate_chain_file: "./cert_chain.pem"
  - type: k8s-cluster
    admin_group_object_ids: ["35959136-eccf-44ec-beed-559a8b33c18e"]
    max_nodes: 6
  - type: k8s-base
```

The byo-cert's module processor then does validation on the inputted certificate before passing the raw string values over to the module (the terraform module itself is empty, it's used as an anchor t fetch the cert texts from). The azure k8s base is now integrated to use the byo-cert if present and we can do a similar approach on the aws and gcp clouds.

Screen Shot 2021-07-22 at 12 32 25 AM


config/registry.yaml
```diff
@@ -121,8 +121,12 @@ modules:
       linkerd_high_availability: optional
       linkerd_enabled: optional
       delegated: optional
-      hosted_zone_name: optional
-    outputs: {}
+      private_key: optional
+      certificate_body: optional
+      certificate_chain: optional
```
legalosLOTR, line 126: Why do we need a cert chain? Isn't that a rare case?

config/registry.yaml
```diff
@@ -308,6 +312,17 @@ modules:
       api_key: optional
       timeout: optional
       values: optional
+  byo-cert:
```
legalosLOTR, line 315: Let's name this "ssl-certificate"? `byo-cert` is not clear enough

opta/module_processors/byo_cert.py
```diff
@@ -0,0 +1,103 @@
...
+
+from opta.exceptions import UserErrors
+from opta.module_processors.base import ModuleProcessor
+
+if TYPE_CHECKING:
+    from opta.layer import Layer
+    from opta.module import Module
+
+
+class BYOCert(ModuleProcessor):
```
legalosLOTR, line 21: Not a big fan of this validation. This is code we need to maintain PLUS there might be edge cases here which lead to false negatives. Like what happens if I don't have a cert chain?

legalosLOTR:
So every time I run `opta apply`, do I need to make sure the cert is there at that exact path? We can't expect users to share the private key file with every dev who needs to run `opta apply`.

config/registry.yaml
juandiegopalomino: I don't think I've ever seen a certificate trusted by a root CA of the internet and unless you're some organization in charge of one of these CAs in which case wow.

opta/module_processors/byo_cert.py
juandiegopalomino: See message above-- unless you're one of the internet heavyhitters this won't be an issue

config/registry.yaml
juandiegopalomino: I thought ssl-certificate was not clear as it would imply we would create it for them lol. How about "external-ssl-cert"?

legalosLOTR: `external-ssl-cert` sounds good!

config/registry.yaml
legalosLOTR: Ah I see. It's fine then.

legalosLOTR (approved):

PR #308, opened 2021-07-28
Adding external ssl support for gcp and aws

juandiegopalomino:
The trickiest was GCP. You see, a gcp http load balancer will only forward 443 requests if there is a gcp ssl attached to it, which suck. I considered provisioning an L4 TCP load balancer just for this case (we don't use it before because it does not handle ssl certs at all-- so with the gcp lbs ssl is either required for port 443 or impossible to handle) and I thought it's way too much of a complication. So in GCP we upload the certificate to GCP, and then also add it to the nginx ingress as the default cert just to keep that behavior consistent thoughout all 3 clouds.


config/tf_modules/k8s-service/k8s-service/templates/ingress.yaml
```diff
@@ -24,14 +24,6 @@ metadata:
     nginx.ingress.kubernetes.io/proxy-send-timeout: "3600"
     {{- end }}
 spec:
-  tls:
```
legalosLOTR, line 27: Would this have any backwards compatibility consequences?

examples/airflow/README.md
```diff
@@ -1,12 +0,0 @@
-# Description
-
```
legalosLOTR, line 2: Why is this being deleted?

config/tf_modules/k8s-service/k8s-service/templates/ingress.yaml
juandiegopalomino: No

examples/airflow/README.md
juandiegopalomino: I thought this was a typo/incorrectly added file. Do you want to keep it?

legalosLOTR: Yup we should keep these files

legalosLOTR (approved):

PR #313, opened 2021-08-04
Better Error Logging for Bucket Uniqueness

nsarupr:


juandiegopalomino (commented):
Great job, one minor comment is that instead of checking the credential config in _fetch_availability_zones, we do it more properly in its own method. Try creating some very short validation function based on this:
https://stackoverflow.com/questions/33332050/getting-the-current-user-account-id-in-boto3

juandiegopalomino (approved):

juandiegopalomino (commented):
whoopsie, don't merge will figure out the testing myself now

PR #316, opened 2021-08-06
feature/RUNX-579-healthchecks-optional

bigbitbus:
Added code to make the healthcheck code optional. This changes the helm charts for the services in the providers, as well as passes the healthcheck_path string into the helm charts.

In the case when the string healthcheck_path is not present in the opta.yaml file, liveliness and readiness probes will not be created for the deployment.


juandiegopalomino (changes requested):
So right now there are 3 variables in play in terraform: healthcheck_path, liveness_probe_path, and readiness_probe_path. With the other two, the healthcheck_path seems redundant but that is precisely the case.

Right now, the default behavior is to use "/healthcheck" for the liveness and readiness probes, but the healthcheck_path is a "shortcode" users can use to override the path for both probes at the same time (cause 99% of the time they really are the same check). The real change is making the default behavior that the path for both probes is NULL and if the helm chart sees that it is so, then it excludes the checks from the generated yaml-- we don't need to introduce healthcheck_path variable into the helm chart.

bigbitbus:
I dont see how introducing a new variable into the helm chart is an issue; I would say the bigger issue I saw in this implementation was that there are 3 different helm charts for each provider in the code, violating the DRY principle. Maybe thats something to invest time in this part of the code.

juandiegopalomino (approved):
awesome!

PR #325, opened 2021-08-17
Record last opta version used and disallow applies from older version

juandiegopalomino:


opta/commands/apply.py
```diff
@@ -135,14 +142,27 @@ def _apply(
...
+    if previous_config is not None and "opta_version" in previous_config:
+        old_opta_version = previous_config["opta_version"]
+        if (
+            old_opta_version != DEV_VERSION
+            and semver.VersionInfo.parse(VERSION.strip("v")).compare(
+                old_opta_version.strip("v")
+            )
+            < 0
+        ):
+            raise UserErrors()
```
legalosLOTR, line 163: Put some helpful message here?

juandiegopalomino: whoops, can't belice I forgot that one! All good now

opta/commands/apply.py
```diff
@@ -160,7 +160,9 @@ def _apply(
             )
             < 0
         ):
-            raise UserErrors()
+            raise UserErrors(
+                f"Can not run with older version-- last run at {previous_config['date']} with version {old_opta_version}. Please update to that version or later"
```
legalosLOTR, line 164: Message could be slightly clarified as `You're trying to run an older version of opta (last run was at <date> with version <version>). Please update to the latest version and try again!`

legalosLOTR (approved):

PR #326, opened 2021-08-18
Deploy Updated Image when deploying with Same Name Tag.

nsarupr:
### Feature
Deploying the Image on the basis of Image Digest, when Image is deployed with the same tag already Running. 

### Affected Functionality on Opta
Deploy

### Explanation
If an Image with **Tag: XYZ** is already deployed, **deploying another image** with the same **Tag: XYZ** will not make any changes to the Infrastructure if the **images are identified using only the _Tag_**.
With this change, **Digest**(_unique for every image generated_) would be taken into account, and Terraform would take Digest as an input which would lead to deploying the Image with the **Same Tag/Different Digest**


opta/commands/deploy.py
```diff
@@ -4,7 +4,7 @@
 
 from opta.amplitude import amplitude_client
 from opta.commands.apply import _apply
-from opta.commands.push import _push, get_push_tag, is_service_config
+from opta.commands.push import _push, is_service_config
```
juandiegopalomino, line 7: I know it was already there, but functions that are to be exported should not start with an underscore-- pls change its name to push

opta/commands/push.py
```diff
@@ -102,7 +129,10 @@ def push_to_docker(
         check=True,
     )
     nice_run(["docker", "tag", local_image, remote_image_name], check=True)
-    nice_run(["docker", "push", remote_image_name], check=True)
+    push_record = nice_run(
```
juandiegopalomino, line 132: my only question is what happens during the docker push command now? Right now we can see the docker push "working" in the cli with those nifty "layer upload progress bars". Are those gone now?

nsarupr: It still prints those elements, but I am checking the Digest based on the Image tag that is being uploaded.

opta/commands/deploy.py
nsarupr: There already exists a `push` method present.

juandiegopalomino (approved):

PR #332, opened 2021-09-01
Mute Old Pods Logs

nsarupr:


opta/core/kubernetes.py
```diff
@@ -343,13 +343,18 @@ def tail_module_log(
...
+    """Using the UTC Time stamp as the Kubernetes uses the UTC Timestamps."""
+    start_time = datetime.datetime.utcnow()
     for event in watch.stream(
         v1.list_namespaced_pod,
         namespace=layer.name,
         label_selector=f"app.kubernetes.io/instance={layer.name}-{module_name}",
     ):
         pod: V1Pod = event["object"]
         color_idx = count % (256 - start_color_idx) + start_color_idx
+        if compare_datetime(start_time, pod.metadata.creation_timestamp) == -1:
```
juandiegopalomino, line 355: are you sure pod.metadata.creation_timestamp has timezone?

opta/utils.py
```diff
@@ -172,3 +173,21 @@ def check_opta_file_exists(config_path: str) -> Literal[True]:
...
+    """
+    Compares date_time2 w.r.t. date_time1.
+    Note: Replacing the TZ Info of date_time1 with TZ Info of date_time2.
+
+    Returns:
+        int:    0:  when date_time1 is equal to date_time2
+                1:  when date_time1 is less than date_time2
+                -1: when date_time1 is greater than date_time2
+    """
+    date_time1.replace(tzinfo=date_time2.tzinfo)
```
juandiegopalomino, line 188: you can't compare 2 times with different timezones?

opta/utils.py
```diff
@@ -172,3 +173,21 @@ def check_opta_file_exists(config_path: str) -> Literal[True]:
             )
         )
     return True
+
+
+def compare_datetime(date_time1: datetime, date_time2: datetime) -> int:
```
juandiegopalomino, line 178: why are you making this function? It's 2 lines which we only use once and you're adding extra features which we never use.

opta/core/kubernetes.py
nsarupr: They have a default Timezone of UTC0. But it can be changed. Let me check on how I can get the timezone of the Pod.

opta/utils.py
nsarupr: Looked cleaner and helped with the test cases as well. Will try to remove it and update the test cases as well.

nsarupr: Removed

opta/utils.py
nsarupr: method removed.

juandiegopalomino (approved):

PR #334, opened 2021-09-02
Prompt when env not supplied

nsarupr:


opta/layer.py
```diff
@@ -220,14 +221,22 @@ def load_from_dict(cls, conf: Dict[Any, Any], env: Optional[str]) -> Layer:
                     )
                 potential_envs[env_name] = (current_parent, env_meta)
 
-            if len(potential_envs) > 1 and env not in potential_envs:
+            if env is None:
+                if len(potential_envs) == 1:
+                    env = list(potential_envs.keys())[0]
+                else:
+                    """This is a repeatable prompt, which will not disappear until a valid choice is provided or SIGABRT
```
juandiegopalomino, line 228: you tested this out?

nsarupr: The test cases with only one Environment passed.
I'll be adding a few more test cases to it.

juandiegopalomino: what I meant did you run this locally, and verified that it keeps prompting you until a valid answer is given?

nsarupr: Yes, it keeps prompting until you enter a valid answer.

nsarupr: Also, I have removed the default value to accidentally avoid a default value.

juandiegopalomino (approved):

PR #336, opened 2021-09-03
Slack Message One Time and Error Prompts

nsarupr:


opta/one_time.py
```diff
@@ -0,0 +1,28 @@
...
+        sys.exit(1)
+
+    print(
+        "\nHi there, thanks for Using Opta.\nWe hope you have a great experience using Opta."
+    )
+
+    slack_prompt()
+
+
+def slack_prompt() -> None:
```
juandiegopalomino, line 27: why make a function that is only used once with a single line?

nsarupr: If some other functionalities are required to be done, they can all be done in the method `one_time`. 
In order to not have a very cluttered method, having single line one time functions made sense.

nsarupr: Removed as per request.

juandiegopalomino (approved):

PR #339, opened 2021-09-07
Fix the Current Image Tag/Digest.

nsarupr:
Fix for PR: #326. 

> After using **_opta deploy_**, doing an **_opta apply_** was an issue with AWS, as the Image would be deployed with an Image Digest, and that was incorrectly mapped as Image Tag due to how `current_image_tag` was configured, which was causing Image Not found exception in Terraform.
> 
> **Image Tag**(`None`) and **Image Digest**(`Actual Value`) would be provided to `_apply` in **_opta deploy_**, as that would help skip the _User check to Kill the Currently running pods_.
> 
> Also, when **_opta apply_** is used, we are checking if an image is already running or not based on the Tag/Digest. This way, we can control whether we want to Kill the existing Pod or not, irrespective of whether the Kubernetes Image contains Image Tag or Image Digest.


juandiegopalomino (commented):
delete current_image_tag as we should not be using it anywhere now

juandiegopalomino (approved):

PR #346, opened 2021-09-13
Adding sticky sessions thanks to nginx ingress

juandiegopalomino:
A k8s-service module spec now has 2 new fields to specify whether to have sticky sessions and how long should the session's cookies last.

### DIFF
--- a/config/registry/aws/modules/aws-k8s-service.yaml
+++ b/config/registry/aws/modules/aws-k8s-service.yaml
@@ -105,6 +105,16 @@ inputs:
     validator: str(required=False)
     description: Use if liveness probe != readiness probe
     default: null
+  - name: sticky_session
+    user_facing: true
+    validator: bool(required=False)
+    description: Use [sticky sessions](https://stackoverflow.com/questions/10494431/sticky-and-non-sticky-sessions) via cookies for your service (first request will send you a cookie called opta_cookie which you should add on future requests).
+    default: false
+  - name: sticky_session_max_age
+    user_facing: true
+    validator: int(required=False)
+    description: If the sticky session is enabled, how long should the cookie last?
+    default: 86400
   - name: resource_request
     user_facing: true
     validator: any(required=False)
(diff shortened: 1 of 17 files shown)


NitinAgg (approved):

config/registry/aws/modules/aws-k8s-service.yaml
```diff
@@ -105,6 +105,16 @@ inputs:
...
+  - name: sticky_session
+    user_facing: true
+    validator: bool(required=False)
+    description: Use [sticky sessions](https://stackoverflow.com/questions/10494431/sticky-and-non-sticky-sessions) via cookies for your service (first request will send you a cookie called opta_cookie which you should add on future requests).
+    default: false
+  - name: sticky_session_max_age
+    user_facing: true
+    validator: int(required=False)
+    description: If the sticky session is enabled, how long should the cookie last?
+    default: 86400
```
bigbitbus, line 117: Is the cookie renewed, or is it expired? From the canvas point of view it may be an issue if suddenly everybody gets booted out of the joint cookie session...

bigbitbus (commented):
Looks great...

How did we test this?

juandiegopalomino:
@bigbitbus forgive me, I forgot to add how I tested this:

I added sticky session in the http example service and opta applied successfully. This new deployment has stdout logs of requests enabled and I created 2 pods for it. From their I hit one of the endpoints from my browser and got expected, successful, response back along with the new opta-cookie. Hitting it again from the browser and I keep getting successful responses with the same cookie value, and I see logs from the **same** pod. I open up a postman session and made a request to the same endpoint with the original cookie and its value and got a successful reploy from the same pod. I repeat in the browser and postman a couple dozen times--always same pod. I open up an incognito window and try the same request-- and we got logs from the other pod.

config/registry/aws/modules/aws-k8s-service.yaml
juandiegopalomino: great question-- a new cookie is created and silently inserted. Canvas has the cookie expiration time so they should be able to work with a graceful replace.

PR #350, opened 2021-09-14
Add support for sh

nsarupr:


opta/commands/shell.py
```diff
@@ -49,7 +58,7 @@ def shell(env: Optional[str], config: str) -> None:
             pod_list[0].metadata.name,
             "-it",
             "--",
-            "bash",
+            use_shell,
```
juandiegopalomino, line 61: what's the use_shell command? Don't you mean sh?

nsarupr: It's not a command, but a variable which will have values from command line. I meant to use shell as the variable but that was already the definition name.

juandiegopalomino (approved):

opta/commands/shell.py
NitinAgg: I agree with @juandiegopalomino --use-shell is a poor flag name in this case. Let's make this --type @nsarupr

PR #357, opened 2021-09-21
Enable consistent hashing

wpride:
The Canvas team would like to enable [consistent hashing](https://www.nginx.com/resources/wiki/modules/consistent_hash/) for our `nginx` ingress so that we can route requests to instances based on the query parameter in the URL.

I tried to follow the pattern [here](https://github.com/run-x/opta/pull/346) since I think functionally this should be identical. The tests pass; however, I haven't been able to test this out since I'm getting this error running locally:

```
(opta) ➜  opta git:(wsp/consistent-hash) opta apply -c ~/sheets/opta/canvas-server/opta.yml --env staging
Opta file validation failed with errors:
  : modules.0: consistent_hash: Unexpected element
```

I'm just running `make build-binary` - did I miss a file, or do I need to run something else? Thank you!


config/version.txt
```diff
@@ -1 +1 @@
-dev
```
juandiegopalomino, line 1: ?

wpride: This must've happened automatically? will remove

juandiegopalomino (approved):

PR #359, opened 2021-09-21
Making sure order does not matter for gcp dns in k8s base

juandiegopalomino:
Tried locally once, now destroying and recreating to confirm consistency, but ready for review

### DIFF
--- a/config/registry/google/modules/gcp-k8s-base.yaml
+++ b/config/registry/google/modules/gcp-k8s-base.yaml
@@ -19,7 +19,7 @@ inputs:
     default: false
   - name: hosted_zone_name
     user_facing: false
-    description: Name of hostezed zone to setup the ingress with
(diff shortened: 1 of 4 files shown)


config/registry/google/modules/gcp-k8s-base.yaml
```diff
@@ -18,6 +18,10 @@ inputs:
     description: The arn for the open id for the eks cluster.
     default: false
   - name: hosted_zone_name
+    user_facing: false
+    description: Name of hostezed zone to setup the ingress with
```
bigbitbus, line 22: hosted

bigbitbus (commented):
Is the change backward compatible (for example, a user who has a opta.yml file for GCP service)?

juandiegopalomino:
@bigbitbus yup, tested it against my existing gcp env and there were no problems

PR #367, opened 2021-09-23
Jd/stop infinite cycle

juandiegopalomino:
# Description
* If a env has the same name as a service, it will raise an error
* If a parent has a parent we raise an issue

# Safety checklist

I ran opta apply locally against a dummy aws app and there was no problem

### DIFF
--- a/opta/layer.py
+++ b/opta/layer.py
@@ -142,7 +142,9 @@ class Layer:
                 )
 
     @classmethod
-    def load_from_yaml(cls, config: str, env: Optional[str]) -> Layer:
+    def load_from_yaml(
+        cls, config: str, env: Optional[str], is_parent: bool = False
+    ) -> Layer:
         t = None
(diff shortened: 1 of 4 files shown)


opta/layer.py
```diff
@@ -189,14 +191,20 @@ def structured_config(self) -> StructuredConfig:
...
         modules_data = conf.get("modules", [])
         environments = conf.pop("environments", None)
         original_spec = conf.pop("original_spec", "")
         path = conf["path"]
         name = conf.pop("name", None)
         if name is None:
             raise UserErrors("Config must have name")
+        if is_parent and environments is not None:
+            raise UserErrors(
+                f"Layer {name} can not be a parent as it has a parent of its own"
```
NitinAgg, line 206: Let's not use "Layer" and "parent" for user facing stuff. But environment and app/service.

NitinAgg (approved):

PR #368, opened 2021-09-23
Update README to be clearer and more welcoming

legalosLOTR:
# Description
Made some minor changes to the readme

# Safety checklist

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,14 +33,22 @@ Supercharge DevOps on any cloud</p>
   </a>
   </p>
 
-# Introduction
-Opta is a platform for running containerized workloads on the cloud. It
-abstracts away the complexity of networking, IAM, kubernetes, and various other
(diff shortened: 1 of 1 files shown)


NitinAgg (approved):

juandiegopalomino (approved):
lgtm! Maybe we want to add something like a [contributions](https://github.com/prometheus/prometheus#contributing) or [community resources](https://github.com/flyteorg/flyte#----community--resources)  section (can be done in the future)?

legalosLOTR:
> lgtm! Maybe we want to add something like a [contributions](https://github.com/prometheus/prometheus#contributing) or [community resources](https://github.com/flyteorg/flyte#----community--resources) section (can be done in the future)?

Will do it in a future version. Not the highest priority right now.

PR #372, opened 2021-09-23
Added docs about postgres backups and upped aws retention from 5->7 days

juandiegopalomino:
# Description
Apparently we had 5 days retention for AWS. Increasing to 7 does not cause data loss or downtime (as it just means keep stuff for longer)

# Safety checklist

opta applied test aws service

### DIFF
--- a/config/registry/aws/modules/aws-postgres.md
+++ b/config/registry/aws/modules/aws-postgres.md
@@ -11,6 +11,12 @@ This module creates a postgres Aurora RDS database instance. It is made in the
 private subnets automatically created during environment setup and so can only be accessed in the
 VPC or through some proxy (e.g. VPN).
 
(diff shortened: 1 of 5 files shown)


config/registry/aws/modules/aws-postgres.md
```diff
@@ -11,6 +11,12 @@ This module creates a postgres Aurora RDS database instance. It is made in the
 private subnets automatically created during environment setup and so can only be accessed in the
 VPC or through some proxy (e.g. VPN).
 
+### Backups
+Opta will provision your database with 5 days of automatic daily backups in the form of
```
NitinAgg, line 15: 7 now?

NitinAgg (approved):
Small comment

PR #373, opened 2021-09-24
Opta Local Changes

bigbitbus:
# Description

Changes to integrate Opta local

# Safety checklist

manual testing, tests tbd.


config/tf_modules/local-base/linkerd.tf
```diff
@@ -0,0 +1,88 @@
...
+  allowed_uses = [
+    "crl_signing",
+    "cert_signing",
+    "server_auth",
+    "client_auth"
+  ]
+}
+
+
+resource "helm_release" "linkerd" {
```
juandiegopalomino, line 57: so we're installing linkerd for them? I thought we decided it was too heavy and not useful for local workloads lol

juandiegopalomino (approved):
one tiny question, otherwise lgtm

config/tf_modules/local-base/linkerd.tf
bigbitbus: Its disabled by default (does not install because of the flag)

juandiegopalomino: ok cool

PR #376, opened 2021-09-27
Added retention limit & snapshot window.

nsarupr:
# Description
Added an option for the user to add the retention limit and a snapshot window for the users to customise the Redis Cache.
`Note: Snapshot window by itself doesn't do anything. It needs retention limit to be set for it to work.`

# Safety checklist

Tested Manually by applying the changes multiple times. Neither was the data lost nor was the pod killed.

### DIFF
--- a/config/registry/aws/modules/aws-redis.yaml
+++ b/config/registry/aws/modules/aws-redis.yaml
@@ -32,7 +32,7 @@ inputs:
     user_facing: true
     validator: str(required=False)
     description: Window during which the snapshot would be taken.
-    default: ""
+    default: "04:00-05:00"
 outputs:
   - name: cache_host
     export: true
--- a/config/registry/aws/modules/aws-redis.yaml
+++ b/config/registry/aws/modules/aws-redis.yaml
@@ -23,6 +23,16 @@ inputs:
     validator: str(required=False)
     description: the redis version to use for this instance
     default: "6.x"
+  - name: snapshot_retention_limit
+    user_facing: true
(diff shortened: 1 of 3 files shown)


config/registry/aws/modules/aws-redis.yaml
```diff
@@ -23,6 +23,16 @@ inputs:
...
+  - name: snapshot_retention_limit
+    user_facing: true
+    validator: int(required=False)
+    description: Retentions days for which the snapshot will be retained.
+    default: 0
+  - name: snapshot_window
+    user_facing: true
+    validator: str(required=False)
+    description: Window during which the snapshot would be taken.
+    default: ""
```
juandiegopalomino, line 35: default: "04:00-05:00"

juandiegopalomino (approved):

config/registry/aws/modules/aws-redis.yaml
```diff
@@ -23,6 +23,16 @@ inputs:
...
     default: "6.x"
+  - name: snapshot_retention_limit
+    user_facing: true
+    validator: int(required=False)
+    description: Retentions days for which the snapshot will be retained.
+    default: 0
+  - name: snapshot_window
+    user_facing: true
+    validator: str(required=False)
+    description: Window during which the snapshot would be taken.
```
NitinAgg, line 34: Can you mention UTC here

NitinAgg (commented):
Sorry for commenting after the merge. But very small fix thanks

PR #389, opened 2021-09-29
Attempting more graceful usage of git dependency

juandiegopalomino:
# Description
Import is not within a try-catch and only imported if the code for its usage is invoked.

# Safety checklist

n/a


bigbitbus (approved):
lgtm

juandiegopalomino:
closing as I need to make it on develop branch

PR #391, opened 2021-09-30
Nicer plan displayer

juandiegopalomino:
# Description
1. Using tables to show data about changes
2. Only 2 security levels "LOW RISK" and "HIGH RISK". Low risk represents no expected dataloss and minimal, recoverable downtime

# Safety checklist

I ran it a series of times locally against opta yaml in different states

Here is what a new run looks like
Screen Shot 2021-09-29 at 5 13 58 PM

Here is what no changes looks like
Screen Shot 2021-09-29 at 5 12 35 PM

Here is what it looks like when I modified the max node count (also called the lb to update itself)
Screen Shot 2021-09-29 at 5 09 42 PM

Here is when I did a normal deploy for an app and also a risky change for a helm chart
Screen Shot 2021-09-29 at 5 00 51 PM

### DIFF
--- a/opta/core/plan_displayer.py
+++ b/opta/core/plan_displayer.py
@@ -172,7 +172,7 @@ class PlanDisplayer:
                 current_risk = resource_change["risk"]
                 table.append(
                     [
-                        module_name,
+                        f"{fg('blue')}{module_name}{attr(0)}",
                         resource_name,
                         resource_change["action"],
                         f"{RISK_COLORS[current_risk]}{current_risk}{attr(0)}",
--- a/opta/core/plan_displayer.py
+++ b/opta/core/plan_displayer.py
@@ -57,7 +57,10 @@ def dict_diffs(dict1: Dict, dict2: dict) -> Dict[Any, Tuple[Any, Any]]:
 class PlanDisplayer:
     @staticmethod
     def handle_update(resource_change: Dict) -> Tuple[str, str]:
-
+        # TODO: use OPA
+        diff_dict = dict_diffs(
+            resource_change["change"]["before"], resource_change["change"]["after"]
+        )
         if resource_change.get("change", {}).get("resource_change", {}) != {}:
             return LOW_RISK, "refreshing data"
         if (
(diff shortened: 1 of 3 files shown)


NitinAgg:
cc @legalosLOTR

juandiegopalomino:
Screen Shot 2021-09-29 at 6 46 34 PM
fyi, first column is now blue

NitinAgg:
Increasing hte nodes/containers should not be high severity. decreasing maybe..

juandiegopalomino:
> Increasing hte nodes/containers should not be high severity. decreasing maybe..

now that is the kind of bespoke logic we can add here!

opta/core/plan_displayer.py
```diff
@@ -1,60 +1,111 @@
...
-        "things instead of changing and deleting things. You should feel very confident in these changes "
-        "without knowing anything further."
-    ),
-    MODERATE_SEVERITY: (
-        "This is the medium level of severity, meaning that something have been updates, but not destroyed or forced "
-        "to replace (destroy and recreate). Typically, this means you updated some minor fields the yaml or updated "
-        "to a new pta version and we're doing some backwards compatible changes."
+RISK_EXPLANATIONS = {
+    LOW_RISK: (
+        "This is the safe level of risk. For these changes there is no expected downtime or dataloss of any sort. "
```
legalosLOTR, line 26: Just write “Low risk change.”

opta/core/plan_displayer.py
```diff
@@ -1,60 +1,111 @@
...
+        "You should feel confident in the changes listed here."
     ),
-    SERIOUS_SEVERITY: (
-        "This is the highest level of severity and this means that either you are destroying something or made a big "
-        "change which forced something to recreate (destroy and create itself). Please tread carefully and know what "
-        "you are doing."
+    HIGH_RISK: (
+        "This is the high level of risk. For these changes there MAY (but not guaranteed) be some donwtime "
```
legalosLOTR, line 30: Just write “High risk change. May lead to downtime and/or data loss.”

opta/core/plan_displayer.py
```diff
@@ -64,62 +115,80 @@ def display(detailed_plan: bool = False) -> None:
...
             )
-            plan_severity = _max_severity(plan_severity, current_severity)
+            plan_risk = _max_risk(plan_risk, current_risk)
 
         logger.info(
-            f"Identified total severity of {SEVERITY_COLORS[plan_severity]}{plan_severity}{attr(0)}.\n"
-            f"{SEVERITY_EXPLANATIONS[plan_severity]}\n"
+            f"Identified total risk of {RISK_COLORS[plan_risk]}{plan_risk}{attr(0)}.\n"
+            f"{RISK_EXPLANATIONS[plan_risk]}\n"
             "If you want extra help, please feel free to reach out to the Runx team at https://slack.opta.dev.\n"
```
legalosLOTR, line 156: For additional help, please reach out to the RunX team at https://slack.opta.dev/

opta/core/plan_displayer.py
```diff
@@ -64,62 +115,80 @@ def display(detailed_plan: bool = False) -> None:
...
+        else:
+            print(
+                tabulate(
+                    table,
+                    ["module", "resource", "action", "risk", "reason"],
+                    tablefmt="fancy_grid",
+                )
+            )
         logger.info(
             "This concludes the plan summary. For more detail, please pass the --detailed-plan flag to your opta "
```
legalosLOTR, line 193: For more details, please rerun the command with the —detailed-plan flag

bigbitbus:
Looks good!
Product enhancement idea: Would we need some machinery (e.g. a map) to specifically mark some changes as high if we determine it to be so? For example, lets say for some reason create VPC is deemed high risk; and experts want to override the machine-generated low-risk categorisation...should we think of a manual override?

juandiegopalomino:
@bigbitbus you make an excellent point and the best answer I think it basically use OPA: https://www.openpolicyagent.org/docs/latest/terraform/
This is a much bigger project (and introduces yet another binary) but it gives us an industry standard way for catching changes to our plan @NitinAgg

NitinAgg (approved):
small comment

PR #399, opened 2021-10-04
Do not allow auto approve if we determine high risk plan

juandiegopalomino:
# Description
Do not allow auto approve if we determine high risk plan

# Safety checklist

n/a


bigbitbus (approved):
lgtm

PR #402, opened 2021-10-05
Add yaml syntax highlighting to README

ajcwebdev:
# Description

Add `yaml` after the opening code blocks to make that `yaml` ✨ pop.

# Safety checklist

With my eyeballs.


NitinAgg (approved):
🚀

PR #403, opened 2021-10-05
Adding Secondary Gcp NodePool Opta module

nsarupr:
# Description
Add an Opta module for Integrating a Secondary Gcp Node Pool in addition to an already existing Node Pool.

Sample Module List:
Screenshot 2021-10-05 at 2 17 27 PM
Sample Google Console UI:
Screenshot 2021-10-05 at 2 17 57 PM

# Safety checklist

Need to manually setup and test a service on the above criterias.

### DIFF
--- a/config/registry/google/modules/gcp-nodepool.yaml
+++ b/config/registry/google/modules/gcp-nodepool.yaml
@@ -1,0 +1,70 @@
+halt: true
+environment_module: true
+inputs:
+  - name: env_name
+    user_facing: false
+    description: Opta Environment name
+    default: None
+  - name: layer_name
+    user_facing: false
+    description: Opta Layer name
+    default: None
+  - name: module_name
+    user_facing: false
+    description: Opta Module name
+    default: None
+  - name: node_zone_names
+    user_facing: false
+    description: The names of the zones to put the nodes in.
+    default: null
+  - name: vpc_self_link
+    user_facing: false
+    description: The self link to the vpc to use.
+    default: null
+  - name: private_subnet_self_link
+    user_facing: false
+    description: The self link to the private subnet to use.
+    default: null
+  - name: k8s_master_ipv4_cidr_block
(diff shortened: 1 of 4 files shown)


config/registry/google/modules/gcp-nodepool.yaml
```diff
@@ -0,0 +1,70 @@
+halt: true
```
nsarupr, line 1: @juandiegopalomino should this be set to `false`

config/tf_modules/gcp-nodepool/main.tf
```diff
@@ -0,0 +1,54 @@
...
+  upper   = false
+}
+
+resource "google_service_account" "gke_node" {
+  account_id   = "opta-${var.layer_name}-${random_string.node_pool_hash.result}"
+  display_name = "opta-${var.layer_name}-default-node-pool"
+}
+
+resource "google_container_node_pool" "node_pool" {
+  name               = "opta-${var.layer_name}-secondary"
```
nsarupr, line 14: @juandiegopalomino @NitinAgg 
Should I rename it something else?

config/registry/google/modules/gcp-nodepool.yaml
juandiegopalomino: yup

config/tf_modules/gcp-nodepool/main.tf
juandiegopalomino: module name

juandiegopalomino (commented):
left 1 important comment

config/tf_modules/gcp-nodepool/main.tf
nsarupr: Updated the name to be a layer_name + random_string

juandiegopalomino (approved):

PR #406, opened 2021-10-05
Jd/fixing displayer

juandiegopalomino:
# Description
plan displayer handle delete. It was delete, not destroy

# Safety checklist

ran locally

### DIFF
--- a/opta/core/plan_displayer.py
+++ b/opta/core/plan_displayer.py
@@ -139,7 +139,7 @@ class PlanDisplayer:
             if action in ["read", "create"]:
                 current_risk = LOW_RISK
                 action_reason = "data_refresh" if action == "read" else "creation"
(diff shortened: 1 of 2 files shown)


opta/core/plan_displayer.py
```diff
@@ -139,7 +139,7 @@ def display(detailed_plan: bool = False) -> None:
             if action in ["read", "create"]:
                 current_risk = LOW_RISK
                 action_reason = "data_refresh" if action == "read" else "creation"
-            elif action in ["replace", "destroy"]:
+            elif action in ["replace", "delete"]:
```
NitinAgg, line 142: let's add some unit tests?

nsarupr (approved):
Looks good to me

NitinAgg (approved):

PR #429, opened 2021-10-19
Opta Module Uniqueness check

nsarupr:
# Description
Check and fast fail on duplicate usage of Opta Modules which don't support multiple instances.

# Safety checklist

This change does not affect anything on the Infrastructure level, hence doesn't change the compatibility with this change.


juandiegopalomino:
Every base, k8sbase and K8s cluster should be unique too

PR #453, opened 2021-11-01
Feature/atlas mongo 1

bigbitbus:
# Description

Adding Mongo Atlas as a database for Opta services in AWS and Local.

# Safety checklist

Manual testing


.github/workflows/ci.yml
```diff
@@ -10,6 +10,15 @@ on:
     branches: [main, dev]
 
 jobs:
+  no-pdb:
```
juandiegopalomino, line 13: huh, I could've sworn I already added that

bigbitbus: Was a rebasing artifact, now this change is not in the PR.

PR #456, opened 2021-11-01
Adding a 10 sleep after tfstate s3 bucket creation

juandiegopalomino:
# Description
As sometimes we get race condtitions for the bucket's existence

# Safety checklist

n/a

### DIFF
--- a/opta/core/terraform.py
+++ b/opta/core/terraform.py
@@ -802,6 +802,7 @@ class Terraform:
                     Bucket=bucket_name,
                     CreateBucketConfiguration={"LocationConstraint": region},
                 )
+            time.sleep(10)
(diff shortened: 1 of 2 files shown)


bigbitbus (approved):

opta/core/terraform.py
```diff
@@ -802,6 +802,7 @@ def _create_aws_state_storage(cls, providers: dict) -> None:
                     Bucket=bucket_name,
                     CreateBucketConfiguration={"LocationConstraint": region},
                 )
+            time.sleep(10)
```
legalosLOTR, line 805: Isn’t there a better way to do this? Like polling for the buckets and existence?

PR #458, opened 2021-11-02
Add Extensible Dependency Graph for Modules.

nsarupr:
# Description
Feature to check whether the Module Dependencies are checked and the configuration fails fast if there are any dependencies missing.

# Safety checklist

The change doesn't change anything on the Infrastructure level, but checks the pre-condition before applying them.


juandiegopalomino (approved):

NitinAgg (approved):
Let's add tests

PR #462, opened 2021-11-03
Several improvements on the gcp service account

juandiegopalomino:
# Description
1. For both the k8s service and gcp service account modules a user can include an extra_project_iam_roles field which is a list of project-level roles to include for the selected accounts. NOTE some services like gcs or bigquery have more fine-grained permissions, like bigquery's [Roles applied on the Dataset level](https://cloud.google.com/bigquery/docs/access-control#dataset_level). This does NOT cover those, just the project level.
2. The service account now has an explicit_name field to give it a user-specified name (as opposed to the constructed one by opta)
3. For both the k8s service and gcp service account modules the read and write bucket fields are exposed to the user in case they want to directly add buckets not in opta (and so they could not use links)

# Safety checklist

tested new fields locally

### DIFF
--- a/config/registry/google/modules/gcp-k8s-service.yaml
+++ b/config/registry/google/modules/gcp-k8s-service.yaml
@@ -31,12 +31,14 @@ inputs:
     description: "Manually set secrets"
     default: [ ]
   - name: read_buckets
-    user_facing: false
-    description: Buckets linked for reading
+    user_facing: true
+    description: Buckets to grant read permissions for
+    validator: list(str(), required=False)
     default: []
   - name: write_buckets
-    user_facing: false
-    description: Buckets linked for writing
+    user_facing: true
+    description: Buckets to grant write permissions for
+    validator: list(str(), required=False)
     default: []
   - name: image
     user_facing: true
@@ -152,6 +154,11 @@ inputs:
       A list persistent storages to add to each instance of your service (need to give a `size` which is the size in GB
       for the storage volume to be, and `path` which is the path in the filesystem of each instance to place it under)
     default: []
+  - name: extra_project_iam_roles
+    user_facing: true
+    description: A list of extra project-level iam roles to grant to the service account created for this k8s service
+    validator: list(str(), required=False)
+    default: [ ]
 extra_validators:
   persistent_storage:
     size: int(required=True)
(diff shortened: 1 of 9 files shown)


config/registry/google/modules/gcp-k8s-service.yaml
```diff
@@ -152,6 +154,11 @@ inputs:
       A list persistent storages to add to each instance of your service (need to give a `size` which is the size in GB
       for the storage volume to be, and `path` which is the path in the filesystem of each instance to place it under)
     default: []
+  - name: extra_project_iam_roles
```
legalosLOTR, line 157: Can I create a role outside of opta and grant it specific permissions for a specific resource and then put it here? 
That way it should be possible to achieve granular permissions. No?

config/registry/google/modules/gcp-k8s-service.yaml
```diff
@@ -31,12 +31,14 @@ inputs:
     description: "Manually set secrets"
     default: [ ]
   - name: read_buckets
-    user_facing: false
-    description: Buckets linked for reading
+    user_facing: true
+    description: Buckets to grant read permissions for
```
legalosLOTR, line 35: I don’t think this should be exposed to the user. If the user wants to add permissions for external buckets they should use “extra_project_iam_roles”

config/registry/google/modules/gcp-k8s-service.yaml
legalosLOTR: How about we name this “additional_iam_roles”?

juandiegopalomino: You are confusing gcp roles with aws roles. gcp roles is more like aws iam policies

config/registry/google/modules/gcp-k8s-service.yaml
```diff
@@ -152,6 +152,11 @@ inputs:
       A list persistent storages to add to each instance of your service (need to give a `size` which is the size in GB
       for the storage volume to be, and `path` which is the path in the filesystem of each instance to place it under)
     default: []
+  - name: additional_iam_roles
+    user_facing: true
+    description: A list of extra project-level iam roles to grant to the service account created for this k8s service
```
NitinAgg, line 157: Add an example here

legalosLOTR (approved):

PR #463, opened 2021-11-03
Test latest vs release version on Github CI

nsarupr:
# Description
Test the release and latest version of Opta for Basic Testing.

# Safety checklist

Running the Github CI.


nsarupr:
Got duplicate Commits

PR #464, opened 2021-11-04
Fixing gcp ingress-path health check issue

juandiegopalomino:
# Description

"""
Hey @patrick nitin and I ran into a bug I need help with. Here’s what is happening
In GCP, when we setup a load balancer, the load balancer needs a health check (simple http, sending a get request and expecting a 200) for its backend  i.e. something which nginx ingress must have. Right now, it is set to send to the path /healthz because that’s the default backend which nginx ingress has. Normally, this works fine: the GCP lb health check pings nginx for /healthz (I don’t know what host it uses) while nginx fulfills all the other ingress routes being set. But here is where the bug happens-- if someone sets up an ingress w/o the host set up and not including the healthz path (or fulfilling it) e.g. public_uri: "/hello" , then the /healthz path now returns a 404, the gcp lb health check fails, and the whole lb gets blocked. You can try this out easily, e.g. get a cluster (doesn’t need to be GCP), apply the ingresses as I have set, use kubectl port forward, and see what happens when you try /healthz. Right now I got no idea I’m confident about to solve this issue, so any help would be appreciated
"""

OK so here is my solution:
nginx already has a dedicated port for healthchecks, it is port 10254 and I confirmed that it is live and well in the ingress nginx. If you hit that port w/ path /healthz you'll get the 200 response.

So I exposed that port in the k8s pod, created a service to listen to the nginx pods on that port, and added and set an ingress to forward all requests to /healthz over to it (and yes, it's ok to create the ingress before the nginx controller exists). It's basically the "let's add a health check deployment" solution but recycling the nginx native health check.

I tested it out and it fixed the lb health checks just as expected. The only catch is that /healthz (no host specified) is now "officially" (it was unofficially due to the bug so no real change) reserved for opta use.

# Safety checklist

ran locally

### DIFF
--- a/config/tf_modules/gcp-k8s-base/ingress_nginx.tf
+++ b/config/tf_modules/gcp-k8s-base/ingress_nginx.tf
@@ -10,6 +10,9 @@ resource "helm_release" "ingress-nginx" {
   values = [
     yamlencode({
       controller : {
+        podLabels: {
+          "opta-ingress-healthcheck" : "yes"
+        }
         extraArgs : var.private_key == "" ? {} : { default-ssl-certificate : "ingress-nginx/secret-tls" }
         config : local.config
         podAnnotations : {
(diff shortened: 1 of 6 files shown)


rabbitfang (commented):
Is there a reason we can't use the ingress controllers health check port? Does GCP support using a different port for health checks?

rabbitfang (approved):

PR #465, opened 2021-11-05
Err User when using Opta via Symlinked directories

nsarupr:
# Description
Using Opta from Symlink leaves Path calculations haywire. So we will now he issuing an error to the User when using Opta via the Symlinked directories.

# Safety checklist

Tested manually on MacOC and Linux machines, to throw an exception when using a Symlinked Directory to run Opta.

### DIFF
--- a/opta/cli.py
+++ b/opta/cli.py
@@ -26,7 +26,6 @@ from opta.commands.validate import validate
 from opta.commands.version import version
 from opta.exceptions import UserErrors
 from opta.one_time import one_time
-from opta.pre_check import pre_check
 from opta.upgrade import check_version_upgrade
 from opta.utils import dd_handler, dd_listener, logger
 
@@ -62,7 +61,6 @@ if __name__ == "__main__":
         # However, we should still clean them up before the next command, or
         # else it may interfere with it.
         one_time()
-        pre_check()
         cleanup_files()
         cli()
     except CalledProcessError as e:
--- a/opta/cli.py
+++ b/opta/cli.py
@@ -26,6 +26,7 @@ from opta.commands.validate import validate
 from opta.commands.version import version
 from opta.exceptions import UserErrors
 from opta.one_time import one_time
+from opta.pre_check import pre_check
 from opta.upgrade import check_version_upgrade
 from opta.utils import dd_handler, dd_listener, logger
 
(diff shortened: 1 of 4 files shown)


juandiegopalomino (approved):
And we are certain there is no _reasonable_ way to make this work, right?

opta/pre_check.py
```diff
@@ -0,0 +1,23 @@
...
+
+from opta.exceptions import UserErrors
+from opta.nice_subprocess import nice_run
+
+
+def pre_check() -> None:
+    is_symlink, cwd_path = is_symlinked_path()
+    if is_symlink:
+        raise UserErrors(
+            f"The current directory seems to be a Symbolic Link.\n"
```
NitinAgg, line 12: Also add. "Currently Opta doesn't support running from a symlinked dir. Please run from a non symlinked directory"

opta/pre_check.py
```diff
@@ -0,0 +1,23 @@
+import os
+from typing import Tuple
+
+from opta.exceptions import UserErrors
+from opta.nice_subprocess import nice_run
+
+
+def pre_check() -> None:
```
legalosLOTR, line 8: We shouldn't run this on every command. Let's do it just on apply and deploy.

opta/cli.py
```diff
@@ -61,6 +62,7 @@ def cli() -> None:
         # However, we should still clean them up before the next command, or
         # else it may interfere with it.
         one_time()
+        pre_check()
```
legalosLOTR, line 65: @rabbitfang you can use the same method for your pre-req checks

opta/pre_check.py
nsarupr: Removed it from the main CLI, and added it to apply and destroy commands only.

opta/cli.py
nsarupr: @rabbitfang I have kept the pre_check.py, but have removed the definition from there. Please add the definition as and when required.

opta/commands/destroy.py
```diff
@@ -51,6 +52,7 @@ def destroy(
     """
 
     config = check_opta_file_exists(config)
+    symlink_check()
```
legalosLOTR, line 55: You don't need this in destroy

nsarupr: destroy would also be needing the paths right? Or is that not path dependent.
If not, I'll revert this PR, and create a fresh one..

legalosLOTR: If it's a symlink path it will get caught during the apply. So we can assume destroy won't run into this.
You don't need to revert the PR - just make a new one and remove this line from destroy.

nsarupr: #472

PR #466, opened 2021-11-05
What started as a crash reporter became a log cleanup

juandiegopalomino:
# Description
1. Created the crash reporter which is automatically generated when opta experiences a non-user error. It contains the opta logs, terraform (detailed) plan, the yaml, any parent yaml, and a metadata json about the version/os/etc...
2. The error loggers where not working as expected and plain missing in some areas. Had to update a couple random lines to make things both show error logs properly and go through the full shutdown sequence as expected.
3. Added some color to current log lines to make them stand out more (red for opta errors, magenta for user errors)
4. Removed unused/deprecated rollback command

TODOS:
1. Include the terraform apply output to the logs (we already have a ticket for that)

# Safety checklist

ran it locally multiple times


opta/cli.py
```diff
@@ -63,22 +64,20 @@ def cli() -> None:
...
-            "If you need more help please reach out to the contributors in our slack channel at: https://slack.opta.dev"
+            f"{fg('magenta')}If you need more help please reach out to the contributors in our slack channel at: https://slack.opta.dev{attr(0)}"
         )
         sys.exit(1)
     except Exception as e:
-        logger.exception(e)
+        logger.exception(str(e))
         logger.info(
-            "If you need more help please reach out to the contributors in our slack channel at: https://slack.opta.dev"
+            f"{fg('red')}Unhandled error encountered -- a crash report zipfile has been createded for you. "
```
NitinAgg, line 76: Why a zipfile and not a single text file?

juandiegopalomino: It makes things so much easier for us (debugging)

juandiegopalomino: if it was difficult, I would have just made a single file but it was so simple to do the zip and so nice to see the contents all neatly displayed

juandiegopalomino: if you disagreee I could certainly make it a single file

NitinAgg (approved):

PR #470, opened 2021-11-09
Fixing logs/events command

juandiegopalomino:
# Description
So what's happening?
* If someone does opta logs, then they will get all logs for all pods printed to stdout (and color-coded like before)
* Like before (e.g. not changing desired behavior), someone can do opta logs -s X where X is an int and get the logs from up to X seconds ago and going forwards
* If someone does opta events, then they will get all events for their service's namespaces printed to stdout
* Like before (e.g. not changing desired behavior), someone can do opta events -s X where X is an int and get the events from up to X seconds ago and going forwards
* When doing opta apply, the current time is passed into the thread directly  so even in the multithread we will accurately know what are the new events/pods/logs (no race condition possible)
* Nicer log fetching backoff  (used to be exponential base 2 which explains why we sometimes did not see the logs)

# Safety checklist

Manual testing

### DIFF
--- a/opta/core/kubernetes.py
+++ b/opta/core/kubernetes.py
@@ -5,7 +5,6 @@ from threading import Thread
 from typing import TYPE_CHECKING, Dict, List, Optional, Set, Tuple
 
 import boto3
-import pytz
 from botocore.exceptions import ClientError, NoCredentialsError
 from colored import attr, fg
 from kubernetes.client import (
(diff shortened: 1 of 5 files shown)


NitinAgg (approved):

opta/core/kubernetes.py
```diff
@@ -470,11 +472,11 @@ def tail_pod_log(
                     )
                     return
 
-            if retry_count < 5:
+            if retry_count < 15:
                 print(
                     f"{fg(color_idx)}Couldn't get logs, waiting a bit and retrying{attr(0)}"
                 )
-                time.sleep(1 << retry_count)
+                time.sleep(retry_count)
```
legalosLOTR, line 479: How about we don't do any backoff here? Let's just do `time.sleep(1)`

PR #478, opened 2021-11-10
Add validation for Providers

nsarupr:
# Description
Add validations for providers.

# Safety checklist

Tested manually with Empty providers or cloud creds.

### DIFF
--- a/opta/layer.py
+++ b/opta/layer.py
@@ -640,7 +640,7 @@ def _validate_providers(providers: dict) -> None:
     """
     if providers is None:
         raise UserErrors(
-            "Environment Configuration contains an the Provider Section.\n"
+            "Environment Configuration needs a Provider Section.\n"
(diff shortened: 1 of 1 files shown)


opta/layer.py
```diff
@@ -627,3 +628,33 @@ def verify_cloud_credentials(self) -> None:
...
+    """
+    Validates Configuration and throws Exception when providers section is provided but left Empty
+    name: Test Name
+    org_name: Test Org Name
+    providers:
+    modules:...
+    """
+    if providers is None:
+        raise UserErrors(
+            "Environment Configuration contains an the Provider Section.\n"
```
NitinAgg, line 643:
```suggestion
            "Environment Configuration needs a Provider Section.\n"
```

NitinAgg (approved):

PR #480, opened 2021-11-10
Service mesh test for GCP

nsarupr:
# Description
Add E2E Service mesh test for GCP

# Safety checklist

YOUR_ANSWER


NitinAgg (approved):
As a next step, add both of these to nightly

nsarupr:
> As a next step, add both of these to nightly

That's the next step.

PR #483, opened 2021-11-12
Check for local PATH dependencies before running a layer

rabbitfang:
# Description
- When running `apply`, `deploy`, or `destroy`, check for PATH dependencies.
- Module processors can override the ModuleProcessor.required_path_dependencies property to return a frozenset of required PATH dependencies.
- Dependencies must be registered (optionally along with the URL to installation instructions) in order to be listed as required (this reduces issues with typos in the code.
- This mechanism allows us to only prompt users for dependencies they actually would need based what modules are present in the layer.
- Updated the Terraform version check to run during deploy and destroy as well.

# Safety checklist

Done some manual testing with a layer file that uses AWS. Other clouds haven't been manually tested yet.

### DIFF
--- a/opta/core/helm.py
+++ b/opta/core/helm.py
@@ -1,11 +1,18 @@
 from subprocess import CalledProcessError  # nosec
-from typing import List
+from typing import FrozenSet, List
 
 from opta.exceptions import UserErrors
 from opta.nice_subprocess import nice_run
 from opta.utils import is_tool, json
+from opta.utils.dependencies import register_path_executable
 
 HELM_INSTALL_URL = "https://helm.sh/docs/intro/install/"
+
+register_path_executable("helm", install_url=HELM_INSTALL_URL)
+
+
+def get_required_path_executables() -> FrozenSet[str]:
+    return frozenset({"helm"})
 
 
 def helm_check() -> None:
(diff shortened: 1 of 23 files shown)


opta/core/kubernetes.py
```diff
@@ -44,6 +45,22 @@
 GCP_CLI_INSTALL_URL = "https://cloud.google.com/sdk/docs/install"
 
 
+register_path_executable("kubectl", install_url=KUBECTL_INSTALL_URL)
+register_path_executable("aws", install_url=AWS_CLI_INSTALL_URL)
+register_path_executable("gcloud", install_url=GCP_CLI_INSTALL_URL)
+register_path_executable("az")
```
juandiegopalomino, line 51: install url?

opta/core/helm.py
```diff
@@ -1,12 +1,19 @@
...
 from opta.nice_subprocess import nice_run
 from opta.utils import is_tool, json
+from opta.utils.dependencies import register_path_executable
 
 HELM_INSTALL_URL = "https://helm.sh/docs/intro/install/"
 
+register_path_executable("helm", install_url=HELM_INSTALL_URL)
+
+
+def get_required_path_executables() -> FrozenSet[str]:
```
juandiegopalomino, line 14: where is this one used?

opta/core/kubernetes.py
rabbitfang: There wasn't one present originally, but I'll look it up and add it.

juandiegopalomino (approved):

PR #484, opened 2021-11-12
Ci/tests/release tests

nsarupr:
# Description

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/.github/workflows/release-tests.yml
+++ b/.github/workflows/release-tests.yml
@@ -1,0 +1,87 @@
+name: Release Testing
+on: workflow_dispatch
+
+jobs:
+  create-destroy-tests-aws:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - name: Pin terraform version
+        uses: hashicorp/setup-terraform@v1
+        with:
+          terraform_version: 1.0.0
+          terraform_wrapper: false
+      - name: Create and destroy test-service and opta environment for AWS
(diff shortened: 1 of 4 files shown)


.github/workflows/release-tests.yml
```diff
@@ -0,0 +1,101 @@
+name: Release Testing
```
NitinAgg, line 1: So this will trigger all of these jobs in parallel and since all of them use the same aws/gcp account it will cause issues. So for now let's use `depends_on` so that for each cloud we only run one of them at a time. But eventually let's use different environment names etc.

.github/workflows/release-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+          terraform_version: 1.0.0
+          terraform_wrapper: false
+      - name: Create and destroy Opta Example on Local
+        run: ./scripts/dispatch_workflow.sh opta create-and-destroy-local.yml main
+        env:
+          github_token: ${{ secrets.ALL_GITHUB_TOKEN }}
+
+  latest-vs-release-test-aws:
+    needs: create-destroy-tests-aws
+    if: ${{ always() }}
```
NitinAgg, line 49: what does this do?

.github/workflows/release-tests.yml
```diff
@@ -0,0 +1,109 @@
...
+    runs-on: ubuntu-20.04
```
NitinAgg, line 50: And you need to add `needs` here as well, right?

NitinAgg (approved):

PR #485, opened 2021-11-12
Websocket CI Test

nsarupr:
# Description

# Safety checklist

YOUR_ANSWER


NitinAgg (approved):

.github/workflows/websocket-e2e-aws.yml
```diff
@@ -0,0 +1,106 @@
+name: Websocket E2E AWS
```
NitinAgg, line 1: Oh wait, can we just add it to create and destroy. Why create a new environment separately?

nsarupr: Yeah, makes sense. Let me make the required changes and send it out for review again.

NitinAgg (approved):

PR #486, opened 2021-11-13
Fix/mongodb rename

bigbitbus:
# Description
Change all mongo to mongodb atlas, as per request from MongoDB folks.

# Safety checklist

Manual testing


NitinAgg (approved):

config/registry/common/modules/mongodb-atlas.md
```diff
@@ -1,10 +1,10 @@
 ---
-title: "Atlas Mongo"
-linkTitle: "Atlas Mongo"
+title: "Mongdb Atlas"
```
legalosLOTR, line 2: `title` and `linkTitle` need to be the same as the module name: `mongodb-atlas`

config/registry/common/modules/mongodb-atlas.md
```diff
@@ -1,10 +1,10 @@
...
+linkTitle: "Mongdb Atlas"
 date: 2021-10-12
 draft: false
 weight: 1
-description: Creates an Atlas Mongo database instance
+description: Creates an Mongdb Atlas database instance
```
legalosLOTR, line 7: Typo in `Mongdb`

legalosLOTR (commented):
Please make sure to test that the mongodb example that we demo'ed works after this change.

config/registry/common/modules/mongodb-atlas.md
bigbitbus: oh ok, I didnt realize that. will fix

PR #487, opened 2021-11-15
Fix for local broken in 18.1

bigbitbus:
# Description

# Safety checklist
* X[ ] This change will NOT lead to data loss

manual test

### DIFF
--- a/opta/module_processors/local_k8s_service.py
+++ b/opta/module_processors/local_k8s_service.py
@@ -234,7 +234,7 @@ class LocalK8sServiceProcessor(LocalK8sModuleProcessor, K8sServiceModuleProcesso
                 "I don't think is what you're looking for."
             )
 
-    def _process_ports(self, data: Dict[Any, Any]) -> None:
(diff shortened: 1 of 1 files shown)


opta/module_processors/local_k8s_service.py
```diff
@@ -234,7 +234,7 @@ def handle_redis_link(
...
             )
 
-    def _process_ports(self, data: Dict[Any, Any]) -> None:
-        # Disable the mulitple ports processing
-        # TODO(patrick): remove once https://github.com/run-x/opta/pull/434 is merged
-        pass
+    # def _process_ports(self, data: Dict[Any, Any]) -> None:
+    #     # Disable the mulitple ports processing
+    #     # TODO(patrick): remove once https://github.com/run-x/opta/pull/434 is merged
+    #     pass
```
rabbitfang, line 240:
```suggestion
```

rabbitfang (approved):
Would prefer to just remove the code instead of commenting it out.

PR #489, opened 2021-11-15
Opta Module Input Schemas

QuinnNgo:
# Description
This PR adds [`JSONSchema`](https://json-schema.org/)s to the Opta repository. These will be served from our GraphQL Gateway for consumption by our frontend.

## TODO
The following is a list of related items that will be included in future work, but are not implemented in this PR.
- Modify `opta validate` to use a module's JSON Schema when there is one
- Add tests to ensure that any `user_facing` fields are included in a module's JSON schema

# Safety checklist

Since this will initially only be consumed by our GraphQL Gateway, tests will be in https://github.com/run-x/graphql-gateway


config/registry/google/modules/gcp-gke.yaml
```diff
@@ -53,7 +53,7 @@ inputs:
   - name: gke_channel
     user_facing: true
     validator: str(required=False)
-    description: The GKE K8s [release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels) to bind the cluster too. Gives you automatic K8s version management for the lcuster and node pools.
+    description: The GKE K8s [release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels) to bind the cluster too. Gives you automatic K8s version management for the  and node pools.
```
rabbitfang, line 56:
```suggestion
    description: The GKE K8s [release channel](https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels) to bind the cluster too. Gives you automatic K8s version management for the cluster and node pools.
```

Is this a typo?

rabbitfang (commented):
How are these generated? This looks like it will be pretty tedious to keep the json scheme in sync with the yamle spec if its all done manually.

QuinnNgo:
> How are these generated? This looks like it will be pretty tedious to keep the json scheme in sync with the yamle spec if its all done manually.

The current state (where both JSON schema and Yamale are in the codebase) is meant to be temporary -- this PR is the first step in a complete migration from Yamale to JSON schema. The eventual vision is that the entire `inputs` array in each module description is deprecated and is completely replaced by its corresponding json schema, and for that schema to be the single source of truth for both validation and docs generation

QuinnNgo:
@rabbitfang Just added a script that takes these descriptions from our existing files and makes sure they're up to date during CI: opta/json_schema.py.

juandiegopalomino (approved):

config/registry/schemas/common-types/cidr-block.json
```diff
@@ -0,0 +1,6 @@
+{
+  "$id": "https://app.runx.dev/common-types/cidr-block",
```
NitinAgg, line 2: why is this needed?

NitinAgg (approved):

config/registry/schemas/common-types/cidr-block.json
QuinnNgo: These are used when schemas reference each other

PR #490, opened 2021-11-15
Datadog logging for terraform issues

bigbitbus:
# Description

# Safety checklist
* [X]  This change will NOT lead to data loss

TBD

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -61,6 +61,9 @@ docker = "5.0.0"
 tabulate = "0.8.9"
 types-tabulate = "0.8.2"
 subprocess-tee = "*"
+psutil = "*"
+cd = "*"
+importlib-metadata = {path = "."}
 
 [requires]
 python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -60,9 +60,7 @@ dnspython = "2.1.0"
 docker = "5.0.0"
 tabulate = "0.8.9"
 types-tabulate = "0.8.2"
-psutil = "*"
-cd = "*"
-importlib-metadata = {path = "."}
+psutil = "5.8.0"
 
 [requires]
 python_version = "3.8"
--- a/Pipfile
+++ b/Pipfile
@@ -60,6 +60,8 @@ dnspython = "2.1.0"
 docker = "5.0.0"
 tabulate = "0.8.9"
 types-tabulate = "0.8.2"
+subprocess-tee = "*"
+
 [requires]
 python_version = "3.8"
 
--- a/Pipfile
+++ b/Pipfile
@@ -60,7 +60,6 @@ dnspython = "2.1.0"
 docker = "5.0.0"
 tabulate = "0.8.9"
 types-tabulate = "0.8.2"
-subprocess-tee = "*"
 psutil = "*"
 cd = "*"
 importlib-metadata = {path = "."}
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -256,7 +256,7 @@ def _apply(
                         f"-out={TF_PLAN_PATH}",
                         layer=layer,
                         *targets,
-                        quiet=True,
+                        quiet=False,
                     )
                 except CalledProcessError as e:
                     logger.error((e.stderr or b"").decode("utf-8"))
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -259,7 +259,7 @@ def _apply(
                         quiet=False,
                     )
                 except CalledProcessError as e:
-                    logger.error((e.stderr or b"").decode("utf-8"))
+                    logger.error(e.stderr or b"")
                     raise e
                 PlanDisplayer.display(detailed_plan=detailed_plan)
 
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -256,7 +256,7 @@ def _apply(
                         f"-out={TF_PLAN_PATH}",
                         layer=layer,
                         *targets,
-                        quiet=False,
+                        quiet=True,
                     )
                 except CalledProcessError as e:
                     logger.error(e.stderr or b"")
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -259,7 +259,7 @@ def _apply(
                         quiet=True,
                     )
                 except CalledProcessError as e:
-                    logger.error(e.stderr or b"")
+                    logger.error(e.stderr or "")
                     raise e
                 PlanDisplayer.display(detailed_plan=detailed_plan)
 
--- a/opta/nice_subprocess.py
+++ b/opta/nice_subprocess.py
@@ -2,6 +2,7 @@ import logging
 import os
 import signal
 import sys
+import tempfile
 from asyncio import TimeoutError
 from pathlib import Path
 from subprocess import CalledProcessError, CompletedProcess  # nosec
@@ -9,7 +10,7 @@ from traceback import format_exc
 from typing import Optional, Union
 
 import psutil
-import tempfile
+
 from opta.constants import DEV_VERSION, VERSION
 from opta.datadog_logging import DatadogLogHandler
 from opta.utils import ansi_scrub
@@ -65,7 +66,7 @@ def nice_run(  # type: ignore # nosec
         log_to_datadog(
             "Calling subprocess with these arguments:\n" + " ".join(*popenargs), "INFO"
         )
-        
+
         with tempfile.TemporaryFile() as f:
             if input:
                 f.write(input)  # type: ignore
--- a/opta/nice_subprocess.py
+++ b/opta/nice_subprocess.py
@@ -63,6 +63,7 @@ def nice_run(  # type: ignore # nosec
         listargs = list(popenargs)
         listargs[0].insert(0, "exec")
         popenargs = tuple(listargs)
+        log_to_datadog("Calling subprocess with these arguments:\n" + " ".join(*popenargs),"INFO")
         if input:
             with open("/tmp/optainput.tmp", "wb") as f:
                 f.write(input)  # type: ignore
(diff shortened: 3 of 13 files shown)


Pipfile
```diff
@@ -60,6 +60,11 @@ dnspython = "2.1.0"
 docker = "5.0.0"
 tabulate = "0.8.9"
 types-tabulate = "0.8.2"
+subprocess-tee = "*"
+psutil = "*"
+cd = "*"
```
rabbitfang, line 65: Specific versions should be pinned

Pipfile
```diff
@@ -60,6 +60,11 @@ dnspython = "2.1.0"
...
+importlib-metadata = {path = "."}
```
rabbitfang, line 66: Was this meant to be left in?

config/tf_modules/k8s-service/main.tf
```diff
@@ -8,7 +8,7 @@ terraform {
 }
 
 resource "helm_release" "k8s-service" {
-  chart = "${path.module}/../../opta-k8s-service-helm"
+  chart = "${path.module}/../../opta-k8s-service-helmz"
```
rabbitfang:
```suggestion
  chart = "${path.module}/../../opta-k8s-service-helm"
```

This looks like a typo or something that wasn't meant to be part of this PR.

opta/amplitude.py
```diff
@@ -73,7 +73,9 @@ def send_event(
             return
         event_properties = event_properties or {}
         user_properties = user_properties or {}
-        insert_id = "".join(random.choices(string.ascii_letters + string.digits, k=16))
+        insert_id = "".join(
+            random.choices(string.ascii_letters + string.digits, k=16)  # nosec
+        )  # nosec
```
rabbitfang, line 78:
```suggestion
        insert_id = "".join(
            random.SystemRandom().choices(string.ascii_letters + string.digits, k=16)
        )
```

See my comment here: https://github.com/run-x/opta/pull/497#discussion_r752931269. Looks like a change to the security tool we use in CI changed which is why its now complaining about this line.

opta/commands/apply.py
```diff
@@ -289,7 +289,7 @@ def _apply(
                         quiet=True,
                     )
                 except CalledProcessError as e:
-                    logger.error((e.stderr or b"").decode("utf-8"))
+                    logger.error(e.stderr or b"")
```
rabbitfang, line 292:
```suggestion
                    logger.error(e.stderr or "")
```

- Don't want to log a `bytes`
- Do we still want to log an empty string if stderr is empty? Does any useful information get logged in that case?

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
     capture_output: bool = False,
     timeout: Optional[float] = None,
     exit_timeout: Optional[float] = None,
+    tee: Optional[bool] = True,
```
rabbitfang, line 27:
```suggestion
    tee: bool = True,
```

- `Optional[T]` means the value can be `None`, which is not what we want in this case.
- Should this default to `False` instead to preserve the existing behavior? That would mean that this PR wouldn't need to update every caller.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
-    it will be used internally.
-
-    By default, all communication is in bytes, and therefore any "input" should
-    be bytes, and the stdout and stderr will be bytes. If in text mode, any
-    "input" should be a string, and stdout and stderr will be strings decoded
-    according to locale encoding, or by "encoding" if set. Text mode is
-    triggered by setting any of text, encoding, errors or universal_newlines.
-
-    The other arguments are the same as for the Popen constructor.
-    """
```
rabbitfang, line 54: If this docstring is no longer accurate, it should be updated instead of removed.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
-            process.kill()
-            # We don't call process.wait() as .__exit__ does that for us.
-            raise
-        retcode = process.poll()
-        if check and retcode:
-            raise CalledProcessError(retcode, process.args, output=stdout, stderr=stderr)
-    return CompletedProcess(process.args, retcode or 0, stdout, stderr)
+
+    try:
+        Path("/tmp/optainput.tmp").touch()
```
rabbitfang, line 33: Use [`tempfile.NamedTemporaryFile`](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) instead of trying to manage temporary files ourselves. Using the files in this way has at least two bugs:
- The file name is not random, so it may conflict with other instances of opta.
- If an exception is thrown, the file is not cleaned up.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+        if input:
+            with open("/tmp/optainput.tmp", "wb") as f:
+                f.write(input)  # type: ignore
```
rabbitfang: If `input` is a string, it will need to be converted to `bytes` before passing to `f.write` (we shouldn't bypass the type system if we can avoid it)

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+        listargs = list(popenargs)
+        listargs[0].insert(0, "exec")
+        popenargs = tuple(listargs)
+        result = run(
+            *popenargs,
+            input=open("/tmp/optainput.tmp", "rb"),
```
rabbitfang: If we are only needing the temp file in order to pass a file-like object, [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO) should be used instead, as it keeps everything in memory and doesn't require touching the filesystem.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+            **kwargs,
+        )
+        os.remove("/tmp/optainput.tmp")
+    except TimeoutError as exc:
+        print("Timeout while running command")
+        signal_all_child_processes()
+        raise exc
+    except KeyboardInterrupt as k:
+        print("Received keyboard interrupt")
+        # Wait again, now that the child has received SIGINT, too.
```
rabbitfang, line 66: The comment says it is waiting again, but it isn't: it immediately re-raises the exception which will bubble up and terminate the parent process.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+        raise k
+    except CalledProcessError as e:
+        raise e
+    except Exception as e:  # Including KeyboardInterrupt, communicate handled that.
+        raise e
+    return result
+
+
+# def nice_run(  # type: ignore # nosec
```
rabbitfang, line 75: This code should be deleted instead of commented out.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+        )
+        os.remove("/tmp/optainput.tmp")
+    except TimeoutError as exc:
+        print("Timeout while running command")
+        signal_all_child_processes()
+        raise exc
+    except KeyboardInterrupt as k:
+        print("Received keyboard interrupt")
+        # Wait again, now that the child has received SIGINT, too.
+        raise k
```
rabbitfang, line 67:
```suggestion
        raise
```

Use short form of `raise` if we are re-raising the same exception without modification.

opta/nice_subprocess.py
```diff
@@ -22,79 +24,133 @@ def nice_run(  # type: ignore # nosec
...
+    except CalledProcessError as e:
+        raise e
+    except Exception as e:  # Including KeyboardInterrupt, communicate handled that.
+        raise e
```
rabbitfang, line 71:
```suggestion

```

`except` clauses that just raise the same exception are unnecessary.

opta/utils/runtee.py
```diff
@@ -0,0 +1,159 @@
...
+
+if TYPE_CHECKING:
+    CompletedProcess = subprocess.CompletedProcess[Any]  # pylint: disable=E1136
+else:
+    CompletedProcess = subprocess.CompletedProcess
+
+try:
+    from shlex import join  # type: ignore
+except ImportError:
+    from subprocess import list2cmdline as join  # nosec
```
rabbitfang, line 20:
```suggestion
from shlex import join
```

`shlex.join` is part of the standard library in Python 3.8, so there shouldn't be a case where it's not present.

opta/utils/runtee.py
```diff
@@ -0,0 +1,159 @@
+# type: ignore
+# nosec
+"""tee like run implementation."""
+import asyncio
```
rabbitfang, line 4: Was this file found at another source? If so, we probably need to mention that and maybe add license text to the file.

rabbitfang (changes requested):

bigbitbus:
We could do this when Python gets a feature (use readuntil with multi separators instead of readline)

https://github.com/python/cpython/pull/16429

tests/test_nice_subprocess.py
```diff
@@ -20,19 +21,19 @@ def test_echo(self):
...
 
     def test_graceful_timeout_exit(self):
         if os.path.exists(GRACEFUL_TERMINATION_FILE):
             os.remove(GRACEFUL_TERMINATION_FILE)
 
-        with pytest.raises(TimeoutExpired):
+        with pytest.raises(TimeoutError):
             nice_run(["python", SIGNAL_HANDLER_SCRIPT], timeout=3)
-
+        sleep(5)
```
NitinAgg, line 36: remove?

NitinAgg (approved):

PR #491, opened 2021-11-15
Default Values peristence and upgrade

juandiegopalomino:
# Description

## The Issue

Right now, for user inputs each module has a default value for each of them to use if the user does not include one themselves. The problem comes for when we want to update the default value and whether the update is NEEDED or optional. We've already had issue where for users their deployment got messed up due to an "optional" change on the default value. Here is the proposed fix:

## Proposal

In the registry modules, each user input will have a new field called "force_update_default_counter" (if it's not set, then it defaults to 0) like so:

```yaml
halt: false
environment_module: true
inputs:
.
.
.
  - name: max_nodes
    user_facing: true
    validator: any(str(), int(), required=False)
    description: Max number of nodes to allow via autoscaling
    default: 15
    force_update_default_counter: 1
```

After a successful opta apply we already store a metadata json in the state storage bucket— now we will add the defaults used

```json
{
  "opta_version": "v0.18.1",
  "date": "2021-11-12T17:37:27.737326",
  "defaults": {
    "module_x": [
      { "input_name": "max_nodes", "default": 15, "force_update_default_counter": 1 }
    ]
  },
  "original_spec": "environments:\n  - name: gcp-staging\n    path: \"../environments/gcp-env.yml\"\n    variables:\n      max_containers: 2\nname: blah40\nmodules:\n  - type: k8s-service\n    image: kennethreitz/httpbin\n    min_containers: 2\n    max_containers: \"{vars.max_containers}\"\n    liveness_probe_path: \"/get\"\n    readiness_probe_path: \"/get\"\n    port:\n      http: 80\n    public_uri: \"test-service.{parent.domain}\"\n    healthcheck_path: \"/healthcheck\"\n    env_vars:\n      - name: APPENV\n        value: \"{env}\"\n      - name: CLOUD_PROVIDER\n        value: gcp\n      - name: hello\n        value: world5\n    links:\n      - pg\n      - redis\n    persistent_storage:\n      - path: \"/DESIRED/PATH1\"\n        size: 5 # 5 GB\n      - path: \"/DESIRED/PATH2\"\n        size: 10 # 10 GB\n  - name: pg\n    type: postgres\n    safety: false\n  - name: redis\n    type: redis\n"
}
```

So upon every opta apply, when generating main.tf.json and selecting the defaults, we shall compare these default values found in the existing metadata json (i.e. the ones used in the last successful opta apply), and reconcile them with the existing ones, using the older values unless the force_default_counter has been incremented, in which case we use the new one found in the registry yaml. After the reconciliation we apply just like before and upload the new default values used. Rinse and repeat

On the first run, it will behave like before— e.g. if no data is found in the metadata.json, it shrugs, keeps the show going and runs with the current default values )because there is no stored state to tell it otherwise).

# Safety checklist

Here is how I tested this out:
I first ran it once and made sure that nothing in terraform changed. I then downloaded the opta config and verified that the existing defaults where preserved as planned. I then changed one random default (initial_readiness_delay) from 30 seconds, to 25, reran opta and confirmed that since I did NOT change/add the force_update_default_counter field, that the initial_readiness_delay kept its original value of 30 (and applied and confirmed that the remote config was not changed too). I did that a couple of times to be sure, and then I updated force_update_default_counter to 1. Sure enough, on the next apply, the old default of 30 was replaced with the new one of 25, and after the successful apply it was changed in the remote config too. I changed the value back to 30 but because I again did NOT update the force_update_default_counter, it stayed at its current value of 25 (and also in remote config). I then updated force_update_default_counter to 2 and once again it picked up the change to the value, setting it to 30.

### DIFF
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -5,6 +5,7 @@ import boto3
 from botocore.config import Config
 from mypy_boto3_dynamodb import DynamoDBClient
 
+from opta.core.cloud_client import CloudClient
 from opta.exceptions import UserErrors
 from opta.utils import fmt_msg, json, logger
 
@@ -22,10 +23,10 @@ class AwsArn(TypedDict):
     resource_type: Optional[str]
 
 
-class AWS:
+class AWS(CloudClient):
     def __init__(self, layer: "Layer"):
-        self.layer = layer
         self.region = layer.root().providers["aws"]["region"]
+        super(AWS, self).__init__(layer)
 
     def __get_dynamodb(self, dynamodb_table: str) -> DynamoDBClient:
         dynamodb_client: DynamoDBClient = boto3.client(
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -26,7 +26,7 @@ class AwsArn(TypedDict):
 class AWS(CloudClient):
     def __init__(self, layer: "Layer"):
         self.region = layer.root().providers["aws"]["region"]
-        super(AWS, self).__init__(layer)
+        super().__init__(layer)
 
     def __get_dynamodb(self, dynamodb_table: str) -> DynamoDBClient:
         dynamodb_client: DynamoDBClient = boto3.client(
--- a/opta/core/cloud_client.py
+++ b/opta/core/cloud_client.py
@@ -8,17 +8,22 @@ class CloudClient:
     def __init__(self, layer: "Layer"):
         self.layer = layer
 
+    @abstractmethod
     def get_remote_config(self) -> Optional["StructuredConfig"]:
         pass
 
+    @abstractmethod
     def upload_opta_config(self) -> None:
         pass
 
+    @abstractmethod
     def delete_opta_config(self) -> None:
         pass
 
+    @abstractmethod
     def delete_remote_state(self) -> None:
         pass
 
+    @abstractmethod
     def get_terraform_lock_id(self) -> str:
-        pass
+        raise NotImplementedError()
--- a/opta/core/cloud_client.py
+++ b/opta/core/cloud_client.py
@@ -5,7 +5,7 @@ if TYPE_CHECKING:
     from opta.layer import Layer, StructuredConfig
 
 
-class CloudClient:
+class CloudClient(ABC):
     def __init__(self, layer: "Layer"):
         self.layer = layer
 
--- a/opta/core/cloud_client.py
+++ b/opta/core/cloud_client.py
@@ -1,0 +1,24 @@
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from opta.layer import Layer, StructuredConfig
+
+
+class CloudClient:
+    def __init__(self, layer: "Layer"):
+        self.layer = layer
+
+    def get_remote_config(self) -> Optional["StructuredConfig"]:
+        pass
+
+    def upload_opta_config(self) -> None:
+        pass
+
+    def delete_opta_config(self) -> None:
+        pass
+
+    def delete_remote_state(self) -> None:
+        pass
+
+    def get_terraform_lock_id(self) -> str:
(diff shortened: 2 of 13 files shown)


opta/core/cloud_client.py
```diff
@@ -0,0 +1,24 @@
+from typing import TYPE_CHECKING, Optional
+
+if TYPE_CHECKING:
+    from opta.layer import Layer, StructuredConfig
+
+
+class CloudClient:
```
rabbitfang, line 7:
```suggestion
class CloudClient(ABC):
```

- Inheriting from `abc.ABC` (abstract base class) prevents CloudClient from being instantiated directly.
- Funnily enough, I had almost done the same thing for https://github.com/run-x/opta/pull/483 but realized it wasn't necessary for that PR.

opta/core/cloud_client.py
```diff
@@ -0,0 +1,24 @@
+from typing import TYPE_CHECKING, Optional
```
rabbitfang, line 1:
```suggestion
from abc import ABC, abstractmethod
from typing import TYPE_CHECKING, Optional
```

See below

opta/core/cloud_client.py
```diff
@@ -0,0 +1,24 @@
...
+        pass
+
+    def delete_opta_config(self) -> None:
+        pass
+
+    def delete_remote_state(self) -> None:
+        pass
+
+    def get_terraform_lock_id(self) -> str:
+        pass
```
rabbitfang, line 24:
```suggestion
    @abstractmethod
    def get_remote_config(self) -> Optional["StructuredConfig"]:
        pass

    @abstractmethod
    def upload_opta_config(self) -> None:
        pass

    @abstractmethod
    def delete_opta_config(self) -> None:
        pass

    @abstractmethod
    def delete_remote_state(self) -> None:
        pass

    @abstractmethod
    def get_terraform_lock_id(self) -> str:
        raise NotImplementedError()
```

- Flag all the methods as abstract to ensure that subclasses override those methods.
- Raise NotImplementedError in `get_terraform_lock` to appease the type checker

opta/core/aws.py
```diff
@@ -22,10 +23,10 @@ class AwsArn(TypedDict):
     resource_type: Optional[str]
 
 
-class AWS:
+class AWS(CloudClient):
     def __init__(self, layer: "Layer"):
-        self.layer = layer
         self.region = layer.root().providers["aws"]["region"]
+        super(AWS, self).__init__(layer)
```
rabbitfang, line 29:
```suggestion
        super().__init__(layer)
```

We should use Python 3's simplified form of `super` which doesn't require passing in the current class and `self`.

opta/core/gcp.py
```diff
@@ -10,20 +10,22 @@
...
 
-class GCP:
+class GCP(CloudClient):
     project_id: Optional[str] = None
     credentials: Optional[Credentials] = None
 
     def __init__(self, layer: "Layer"):
         self.layer = layer
         self.region = layer.root().providers["google"]["region"]
+        super(GCP, self).__init__(layer)
```
rabbitfang, line 28:
```suggestion
        super().__init__(layer)
```

Ditto from above

opta/core/local.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, layer: "Layer"):
         )
         if not os.path.exists(os.path.dirname(self.config_file_path)):
             os.makedirs(os.path.dirname(self.config_file_path))
+        super(Local, self).__init__(layer)
```
rabbitfang, line 23:
```suggestion

        super().__init__(layer)
```

rabbitfang (changes requested):
Just some initial notes. Still need to do a deeper dive into the logic.

rabbitfang (approved):
Minor changes; otherwise, LGTM

PR #492, opened 2021-11-15
Update the Init functionality

nsarupr:
# Description
Update the Init functionality

# Safety checklist

Does not change any existing functionality.


opta/commands/init.py
```diff
@@ -1,130 +1,16 @@
...
-    pass
-
-
-ENVIRONMENT_TEMPLATES: Dict[str, Template] = {
-    "aws": awsTemplate,
-    "gcp": gcpTemplate,
-    "azure": azureTemplate,
-}
-
+    This command brings to you an interactive way of creating an opta configuration file.
```
NitinAgg, line 9:
```suggestion
    Redirecting to `https://app.runx.dev/yaml-generator`
```

opta/commands/init.py
```diff
@@ -1,130 +1,16 @@
...
-    file by using the command `opta init service <environment_file> <template_name>`.
-
-    In order to see the full list of available templates, you can
-    run `opta init env --help` or `opta init service --help` respectively.
-    """
-    pass
-
-
-ENVIRONMENT_TEMPLATES: Dict[str, Template] = {
-    "aws": awsTemplate,
```
NitinAgg, line 52: Can you also delete all the config files which are pointed here

NitinAgg (changes requested):

opta/commands/init.py
nsarupr: `opta -h` would have this as the result.

`Usage: opta [OPTIONS] COMMAND [ARGS]...

  Welcome to opta, runx's cli!

Options:
  -h, --help  Show this message and exit.

Commands:
  apply              Initialize your environment or service to match the...
  configure-kubectl  Configure the kubectl CLI tool for the given cluster
  deploy             Push your new image to the cloud and deploy it in...
  destroy            Destroy all opta resources from the current config
  force-unlock       Force Unlocks a stuck lock on the current workspace
  init               Redirecting to `https://app.runx.dev/yaml-generator`
  logs               Get stream of logs from your service
  secret             Commands for manipulating secrets for a k8s service
  shell              Get a bash shell into one of the pods in your service`

So, need to have a better input message.
Plus, running `opta init` will not show any messages.

opta/commands/init.py
```diff
@@ -1,130 +1,16 @@
...
+@click.command()
 def init() -> None:
     """
-    This command initializes an opta configuration file. You can create an environment
-    configuration file by using the command `opta init env <cloud-provider>` and a service
-    file by using the command `opta init service <environment_file> <template_name>`.
-
-    In order to see the full list of available templates, you can
-    run `opta init env --help` or `opta init service --help` respectively.
+    This command brings to you an interactive way of creating an opta configuration files.
```
NitinAgg, line 9:
```suggestion
    Opens the interactive UI
```

NitinAgg (approved):

NitinAgg (approved):

PR #496, opened 2021-11-17
Update the CI Test configurations.

nsarupr:
# Description
Update the CI Test configurations.

# Safety checklist

Tested Manually.

### DIFF
--- a/.github/workflows/latest-vs-release-test-aws.yml
+++ b/.github/workflows/latest-vs-release-test-aws.yml
@@ -99,13 +99,13 @@ jobs:
           echo "This was the found secret value: ${secret_value}"
           [[ "$secret_value" =~ .*"foo".* ]] || exit 1
 
-      - name: Set stateful data
-        env:
-          OPTA_DISABLE_REPORTING: true
-        run: |
-          export PARENT_DOMAIN=`/home/runner/.opta/opta output --env ci --config ./examples/ci-tests/create-and-destroy-aws/service.yml | jq -r '."parent.domain"'`
-          export HOST=`/home/runner/.opta/opta output --env ci --config ./examples/ci-tests/create-and-destroy-aws/service.yml | jq -r '.load_balancer_raw_dns'`
-          curl -X POST -L --header "Host: test-service.${PARENT_DOMAIN}" --fail ${HOST}/stateful
+#      - name: Set stateful data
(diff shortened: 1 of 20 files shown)


examples/ci-tests/create-and-destroy-aws/environment.yml
```diff
@@ -1,6 +1,6 @@
 # This opta environment is used by the test-ci job in the runxc repo.
-name: opta-ci
-org_name: runx
+name: awsenv-ci
+org_name: runx-ci
```
NitinAgg, line 3: Can we keep this runx so we have proper accounting in amplitude

.github/workflows/latest-vs-release-test-aws.yml
```diff
@@ -99,13 +99,13 @@ jobs:
...
           [[ "$secret_value" =~ .*"foo".* ]] || exit 1
 
-      - name: Set stateful data
-        env:
-          OPTA_DISABLE_REPORTING: true
-        run: |
-          export PARENT_DOMAIN=`/home/runner/.opta/opta output --env ci --config ./examples/ci-tests/create-and-destroy-aws/service.yml | jq -r '."parent.domain"'`
-          export HOST=`/home/runner/.opta/opta output --env ci --config ./examples/ci-tests/create-and-destroy-aws/service.yml | jq -r '.load_balancer_raw_dns'`
-          curl -X POST -L --header "Host: test-service.${PARENT_DOMAIN}" --fail ${HOST}/stateful
+#      - name: Set stateful data
```
NitinAgg, line 102: Why are we removing this

nsarupr: The Output function doesn't return just the output. 
We have added 
https://github.com/run-x/opta/blob/990cf54f56d29a8a25200757ac2bef666dcba2be/opta/module_processors/aws_k8s_base.py#L49
which configures AWS Kube-Config, which fails the JQ command due to additional statement being printed (Json Parsing fails.).

examples/ci-tests/create-and-destroy-aws/environment.yml
nsarupr: Sure, I'll make this change.

NitinAgg (approved):

PR #497, opened 2021-11-18
Support static sites via S3 + Cloudfront

juandiegopalomino:
# Description
This commit includes changes to allow opta users to deploy and manage static sites via S3 + Cloudfront. This is done via the following changes:

1. There's a new cloudfront-distribution module which is linked to an S3 bucket whose files it will be serving/caching. Comes with a default/404 page (which can be overwritten at the user's pleasure) and optionally support SSL/official site domains.
2. The S3 module now supports a files field with which users can automatically upload files to the bucket with (also detects changes)
3. Both modules can work WITHOUT the aws-base module, but that's still recommended because the base module sets up the logging bucket, which is needed for security audits (but again, not necessary for ease of use)
4. I updated my OS and along the why I did a pipenv update causing the pipfile.lock to update with later versions of the packages used but I did not change the version restrictions (so nothing should break) and I'm now afraid of reverting it lol.

Points 1-3 have extension documentation so pls take a look at that too.

# Safety checklist

Created a whole new env for cloudfront, and applied/modified multiple times to check SSL, setting domain, file upload, URL correctness and so on. Everything is working as expected.


opta/module_processors/aws_cloudfront_distribution.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def process(self, module_idx: int) -> None:
+        aws_base_modules = self.layer.get_module_by_type("aws-base", module_idx)
+        from_parent = False
+        if len(aws_base_modules) == 0 and self.layer.parent is not None:
+            from_parent = True
+            aws_base_modules = self.layer.parent.get_module_by_type("aws-base")
+
+        if len(aws_base_modules) == 0:
+            logger.warn(
```
legalosLOTR, line 28: We should not output this warning. It can be minor note in the docs but shouldn't be printed at runtime.

opta/module_processors/aws_s3.py
```diff
@@ -18,21 +19,28 @@ def __init__(self, module: "Module", layer: "Layer"):
...
         aws_base_modules = self.layer.get_module_by_type("aws-base", module_idx)
+        from_parent = False
         if len(aws_base_modules) == 0 and self.layer.parent is not None:
             from_parent = True
             aws_base_modules = self.layer.parent.get_module_by_type("aws-base")
 
         if len(aws_base_modules) == 0:
-            raise UserErrors(
-                "Did not find the aws-eks module in the layer or the parent layer"
+            logger.warn(
```
legalosLOTR, line 28: Let's not do this.

config/registry/aws/modules/aws-cloudfront-distribution.md
```diff
@@ -0,0 +1,64 @@
...
+If you are ready to start hosting your site with your domain via the cloudfront distribution, then proceed as follows:
+1. Get an [AWS ACM certificate](https://docs.aws.amazon.com/acm/latest/userguide/gs-acm-request-public.html) for your site. 
+   Make sure that you get it in region us-east-1. If you already have one at hand in your account (e.g. from another 
+   active Opta deployment), then feel free to reuse that.
+2. In whatever hosted zone you are using to manage the DNS for your domain (or the parent domain), create a new CNAME 
+   record for the domain you wish to use for cloudfront and point it at the `cloudfront_domain` gotten above.
+3. Set the acm_cert_arn and domains fields in opta accordingly
+4. Opta apply and you're done!
+
+### Bucket Access level
```
legalosLOTR, line 55: I don't think we need this. 99% of the time it's just going to be read. Let's not support different access levels in v1.

config/tf_modules/aws-s3/bucket.tf
```diff
@@ -64,6 +66,61 @@ resource "aws_s3_bucket" "bucket" {
...
+
+resource "aws_cloudfront_origin_access_identity" "read_write" {
+  comment = "For writing to bucket ${var.bucket_name}"
+}
+
+resource "aws_cloudfront_origin_access_identity" "read_write_delete" {
+  comment = "For deleting and writing to bucket ${var.bucket_name}"
+}
+
+data "aws_iam_policy_document" "s3_policy" {
```
legalosLOTR, line 82: Can we move this to the cloudfront module? The cf module should add these perms when it gets linked to s3.

config/tf_modules/aws-s3/files.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_s3_bucket_object" "user_files" {
+  for_each     = local.files_to_upload
+  bucket       = aws_s3_bucket.bucket.id
+  key          = each.key
+  source       = "${var.files}/${each.key}"
+  content_type = lookup(local.mime_types, regex("\\.[^.]+$", each.key), "binary/octet-stream")
```
legalosLOTR, line 6: ooh very cool. This would avoid so many problems for the users!
I've had to debug these mime problems in s3 countless times.

legalosLOTR:
Here is an example of a customer request for cloudfront without s3: https://run-x.atlassian.net/browse/RUNX-519. We should design this module such that it can one day be used to support this non-s3 use case as well (no need to support it right now).

config/tf_modules/aws-s3/bucket.tf
juandiegopalomino: No, we really can not because it has to live in the bucket IAM policy, which is set here.

juandiegopalomino: (and a bucket can not hold 2 policies)

config/tf_modules/aws-cloudfront-distribution/default.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
```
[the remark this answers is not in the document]
juandiegopalomino: I think a weird xml one-- in any case we ought to keep this as
1) Verifies to us and users that cloudfront has been created successfully and it absolutely can read from the s3 bucket
2) Free publicity.

config/registry/aws/modules/cloudfront-distribution.yaml
```diff
@@ -0,0 +1,60 @@
+halt: false
+environment_module: false
+terraform_module: aws-cloudfront-distribution
```
legalosLOTR, line 3: Can we remove the `aws-` from here too?

config/tf_modules/aws-s3/files.tf
legalosLOTR, line 6: What happens for a file that doesn't have an extension? Would this error?

config/registry/aws/modules/aws-s3.md
```diff
@@ -24,6 +24,34 @@ This module creates an S3 bucket for storage purposes. It is created with server
...
+    └── hello3.html
+```
+
+Then Opta would upload 4 files to your S3 bucket, with the S3 keys being `hello2.html`, `hello2.txt`, `hello3.txt` and
+`subdir/hello3.html`.
+
+Opta will also catch any changes to the files on the next `opta apply` and will push updates as needed. Opta supports
+extensive MIME parsing, so it also makes sure to set the content type correctly.
 
 ### Linking
```
legalosLOTR, line 56: We should add here that this module can be linked to cloudfront to serve static sites (link to cloudfront docs)

Also we should document on this page somewhere that we are creating a cloudfront OAI.

config/registry/aws/modules/cloudfront-distribution.md
```diff
@@ -0,0 +1,54 @@
+---
+title: "cloudfront-distribution"
+linkTitle: "cloudfront-distribution"
+date: 2021-11-16
+draft: false
+weight: 1
+description: Set up a cloudfront distribution to serve static files from your s3 bucket
```
legalosLOTR, line 7: This should just say "Set up a cloudfront distribution".

config/registry/aws/modules/cloudfront-distribution.md
```diff
@@ -0,0 +1,54 @@
...
+---
+
+This module sets up a cloudfront distribution for you, tailored towards serving static websites/files from an Opta s3
```
legalosLOTR, line 10: Reword
```
This module sets up a cloudfront distribution. Currently, it's not very useful unless you link it with a s3 module.
```

config/registry/aws/modules/cloudfront-distribution.md
```diff
@@ -0,0 +1,54 @@
...
+title: "cloudfront-distribution"
+linkTitle: "cloudfront-distribution"
+date: 2021-11-16
+draft: false
+weight: 1
+description: Set up a cloudfront distribution to serve static files from your s3 bucket
+---
+
+This module sets up a cloudfront distribution for you, tailored towards serving static websites/files from an Opta s3 
+bucket (currently just for a single S3, but will expand for more complex usage in the future). Now, hosting your
```
legalosLOTR, line 11: Create a new section here "Linking with s3" then add
```
Once you link the cloudfront module with s3, it can be used to serve static files from your s3 bucket. See below for an example.
```

config/tf_modules/aws-cloudfront-distribution/main.tf
```diff
@@ -0,0 +1,92 @@
...
+}
+
+resource "aws_s3_bucket_object" "default_page" {
+  bucket       = data.aws_s3_bucket.current_bucket.id
+  key          = "opta-default-page-${random_string.default-suffix.result}.html"
+  source       = "${path.module}/default.html"
+  content_type = "text/html"
+}
+
+resource "aws_cloudfront_distribution" "s3_distribution" {
```
legalosLOTR, line 17: This resource and the one above it should not be created unless `var.bucket_name` is true. The cloudfront module will basically be useless without linking. But we should follow this to keep the module generic.

config/tf_modules/aws-s3/bucket.tf
```diff
@@ -64,6 +66,27 @@ resource "aws_s3_bucket" "bucket" {
   }
 }
 
+resource "aws_cloudfront_origin_access_identity" "read" {
```
legalosLOTR, line 69: Can you add a comment here: that it's fine adding this as it's just creating something akin to an IAM role.

config/registry/aws/modules/cloudfront-distribution.yaml
```diff
@@ -0,0 +1,60 @@
...
+    default: None
+  - name: module_name
+    user_facing: false
+    description: Opta Module name
+    default: None
+  - name: s3_log_bucket_name
+    user_facing: false
+    description: Name of the access log bucket
+    default: null
+  - name: bucket_name
```
legalosLOTR, line 21: I think bucket_name and origin_access_identity_path should also be user_facing and not required. This way someone can use a bucket that they created outside of opta.

config/tf_modules/aws-cloudfront-distribution/main.tf
```diff
@@ -0,0 +1,92 @@
...
+  enabled             = true
+  is_ipv6_enabled     = true
+  comment             = "Opta managed cloudfront distribution ${var.layer_name}-${var.module_name}"
+  default_root_object = var.default_page_file == null ? aws_s3_bucket_object.default_page.key : var.default_page_file
+
+  dynamic "logging_config" {
+    for_each = var.s3_log_bucket_name == null ? [] : [1]
+    content {
+      include_cookies = true
+      bucket          = var.s3_log_bucket_name
```
legalosLOTR, line 36: I don't think we should re-use the s3 logging bucket in cloudfront.

config/registry/aws/modules/cloudfront-distribution.yaml
juandiegopalomino: preferably no-- right now the terraform modules are all stored in a single subdirectory (to be changed soon with module api), and so the cloud prefixes actually started a a means of keeping them somewhat sorted. For all intents and purposes the module as seen by our code and the user is cloudfront-distribution but its terraform module directory happens to be aws-cloudfront-distribution

config/tf_modules/aws-cloudfront-distribution/main.tf
juandiegopalomino: ... why not? This is the environment-wide audit logging bucket used to store logs from the load balancers, buckets, and anything else that supports sending audits to an s3 bucket.

legalosLOTR: Oh sorry I misunderstood it. It's coming from the parent.. that's fine then!

config/registry/aws/modules/cloudfront-distribution.yaml
legalosLOTR: Ah cool. If there is no user visible component, then it's fine!

config/registry/aws/modules/cloudfront-distribution.md
juandiegopalomino: er... that's incorrect. It FAILS unless it's linked to an S3 bucket

config/registry/aws/modules/cloudfront-distribution.md
juandiegopalomino: done

config/registry/aws/modules/aws-s3.md
juandiegopalomino: done

config/tf_modules/aws-s3/files.tf
juandiegopalomino: good catch

config/tf_modules/aws-s3/bucket.tf
juandiegopalomino: done

config/registry/aws/modules/aws-s3.yaml
```diff
@@ -38,10 +38,17 @@ inputs:
...
-    user_facing: bool
+    user_facing: true
     validator: include('cors', required=False)
     description: A custom cors policy.
     default: null
+  - name: files
+    user_facing: true
+    validator: str(required=False)
+    description: |
+      The path (can be relative) to a directory holding files which you wish to upload to the s3 bucket.
```
rabbitfang, line 49: > The path (can be relative)

Relative to what? The cwd when the `opta` is run? The opta config file? opta itself? We should document what the path is relative to.

rabbitfang: Noticed this is documented in the module markdown doc, but we should document it here too.

config/tf_modules/aws-cloudfront-distribution/variables.tf
```diff
@@ -23,7 +23,7 @@ variable "s3_log_bucket_name" {
 
 variable "default_page_file" {
   type    = string
-  default = null
+  default = "index.html"
```
legalosLOTR, line 26: How does this default interact with the default in the yaml? Does one override the other? Ideally we would have the default in one place only.

config/registry/aws/modules/cloudfront-distribution.yaml
```diff
@@ -19,18 +19,21 @@ inputs:
...
+    validator: str(required=False)
     description: The Cloudfront OAI path to use to access the buckets
     default: ""
   - name: default_page_file
     user_facing: true
     validator: str(required=False)
     description: The name of the existing s3 object in your bucket which will serve as the default page.
-    default: null
+    default: "index.html"
+    force_update_default_counter: 1
```
legalosLOTR, line 36: Do we need to bump this?

legalosLOTR (commented):
Thanks for making all the changes! I just had some minor clarifications. Nothing blocking from my side.

config/tf_modules/aws-s3/mime.json
```diff
@@ -0,0 +1,76 @@
+{
```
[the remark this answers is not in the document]
juandiegopalomino: I wrote it manually by manipulating a CSV

config/registry/aws/modules/cloudfront-distribution.yaml
juandiegopalomino: I did because it was not changing the default on my example because I already ran it with default null

config/tf_modules/aws-cloudfront-distribution/variables.tf
juandiegopalomino: You're right-- this is a useless default which always get overwritten by the default in the yaml, I'll remove them now

opta/module_processors/aws_s3.py
```diff
@@ -18,21 +19,28 @@ def __init__(self, module: "Module", layer: "Layer"):
 
     def process(self, module_idx: int) -> None:
         aws_base_modules = self.layer.get_module_by_type("aws-base", module_idx)
+        from_parent = False
```
legalosLOTR, line 22: Testing code?

config/registry/aws/modules/cloudfront-distribution.yaml
legalosLOTR: Let's delete this? We should not commit testing code.

legalosLOTR (approved):

opta/module_processors/aws_s3.py
juandiegopalomino: no, this was a real bug

rabbitfang (approved):
My notes were covered, but I noticed that total test coverage dropped less than 70% in this PR, which I think will cause main branch CI to start failing if we merge this PR?

PR #498, opened 2021-11-18
Ignore region AWS config

nsarupr:
# Description
Ignore the AWS Region if any is mentioned in the AWS configuration directory. 
Override it with the Region mentioned in the Environment Configuration.

# Safety checklist

Tested Manually.

### DIFF
--- a/opta/commands/destroy.py
+++ b/opta/commands/destroy.py
@@ -5,6 +5,7 @@ from typing import Dict, List, Optional
 import boto3
 import click
 from azure.storage.blob import ContainerClient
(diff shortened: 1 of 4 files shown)


NitinAgg (commented):
Looks good. Did you get a chance to run this on a proper example? Let's do that before merging

nsarupr:
> Looks good. Did you get a chance to run this on a proper example? Let's do that before merging

While testing this, got another issue😂😂

juandiegopalomino (approved):

PR #504, opened 2021-11-23
Modules linked to dbs will now need to wait for db to be fully up

juandiegopalomino:
# Description
Just made the outputs depend on the first db instance to be up and running.

# Safety checklist

locally and yup, it's now waiting for db instance


NitinAgg (approved):
Love this!

PR #506, opened 2021-11-29
Jd/circleci

juandiegopalomino:
# Description
create and destroy pipeline ported over for circleci.
We could also support unit tests + linting if we so wish

# Safety checklist

n/a


examples/ci-tests/create-and-destroy-aws/service-docdb.yml
```diff
@@ -0,0 +1,28 @@
+environments:
```
NitinAgg, line 1: I don't see the benefit of separating config for each storage. Let's just have one yaml with everything

.circleci/config.yml
```diff
@@ -21,6 +705,91 @@ jobs:
 # Invoke jobs via workflows
 # See: https://circleci.com/docs/2.0/configuration-reference/#workflows
 workflows:
-  say-hello-workflow:
+  version: 2
+  run-create-and-destroy:
+    when: << pipeline.parameters.run-create-and-destroy >>
+    jobs: &id001
```
NitinAgg, line 711: is it possible to give it a more human readable name?

NitinAgg (approved):

PR #507, opened 2021-11-30
Feat load balancer improvements

juandiegopalomino:
# Description
Oh my gosh, I can't believe that this worked. Hats off to Nilesh for getting this process started. Basically this pr does the following:

1. Starts using the AWS Load Balancer Controller which allows us to do the following
2. Explicitly name the load balancer created and fetch its data (including its dns) on terraform. Byebye clunky python code
3. Set the ALPN policy on terraform. Byebye clunky python code too

# Safety checklist

So here's the crazy part-- when these changes are executed, it does not destroy the old load balancer, it creates a new one in tandem (if it's a totally new environment it will just create 1 load balancer luckily). The old load balancer will keep working fine (until I think new ec2s are added) although the references will all begin pointing to the new one and moreover if delegated is set to true, the dns will automatically switch to use the new load balancer (although there will be a 5 minute downtime).


config/registry/aws/modules/aws-k8s-base.yaml
```diff
@@ -86,6 +86,9 @@ inputs:
     validator: list(int(min=1, max=65535), required=False)
     description: Which additional TCP ports should have TLS enabled
     default: []
-outputs: []
+outputs:
+  - name: load_balancer_raw_dns
+    export: true
+    description: Blah
```
nsarupr, line 92: Better Description.

config/tf_modules/k8s-base/load_balancer_controller.tf
```diff
@@ -0,0 +1,331 @@
...
+  assume_role_policy = data.aws_iam_policy_document.trust_k8s_openid_alb.json
+  name               = "opta-${var.env_name}-albc"
+}
+
+resource "aws_iam_role_policy_attachment" "load_balancer" {
+  policy_arn = aws_iam_policy.load_balancer.arn
+  role       = aws_iam_role.load_balancer.name
+}
+
+resource "helm_release" "load_balancer" {
```
nsarupr, line 308: should we set some basic resource limit for this?

opta/commands/output.py
```diff
@@ -31,52 +28,26 @@ def output(config: str, env: Optional[str],) -> None:
...
-                    tag_description["ResourceArn"]
-                ]["DNSName"]
-                return current_outputs
-
-        if marker == "":
-            break
+        del current_outputs["parent.load_balancer_raw_dns"]
     logger.info(
         "Could not find load balancer for current environment/service. Please re-run opta apply on the environment file."
     )
```
nsarupr, line 46: We should make this conditional.

nsarupr:
@juandiegopalomino 

`opta output` seems to be giving two Load Balancer URLs.
`
"load_balancer_raw_dns": "opta-awsenv-lb-9000c05e77a088b1.elb.us-east-1.amazonaws.com",
"load_balancer_raw_dns_2": "opta-awsenv-lb-6cd891a41f06f208.elb.us-east-1.amazonaws.com"
`

juandiegopalomino:
@nsarupr I think that you're seeing load_balancer_raw_dns_2 in your state due to some development I was doing in that env which you gave me

legalosLOTR (commented):
Great! Let's make sure we clearly understand how the upgrade process would work and document it - before merging this.

PR #508, opened 2021-11-30
Tentative new file structure for module_api

bigbitbus:
# Description
New directory/file structure of opta repo for module-api changes.

# Safety checklist

YOUR_ANSWER


modules/datadog/datadog_logging.py
```diff
@@ -0,0 +1,84 @@
+import os
```
NitinAgg, line 1: this isn't a module

NitinAgg (commented):
I think the datadog logging changes creeped in here. but otherwise looks good

modules/datadog/datadog_logging.py
bigbitbus: Good catch, fixed.

NitinAgg (approved):

bigbitbus:
Lot of changes, for now registry yaml is still picked up from config/

PR #511, opened 2021-12-01
Making sure k8s service can be put in env

juandiegopalomino:
# Description
Code fixes to make sure k8s services can go in env files directly

# Safety checklist

manually run in existing and new envs

### DIFF
--- a/opta/core/generator.py
+++ b/opta/core/generator.py
@@ -27,6 +27,7 @@ def gen(
     image_tag: Optional[str] = None,
     image_digest: Optional[str] = None,
     test: bool = False,
+    check_image: bool = False,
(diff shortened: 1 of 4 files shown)


opta/core/generator.py
```diff
@@ -27,6 +38,31 @@ def gen(
         current_modules.append(module)
         if not module.halt and module_idx + 1 != total_module_count:
             continue
+        service_modules = layer.get_module_by_type("k8s-service", module_idx)
+        if len(service_modules) > 0 and (get_cluster_name(layer.root()) is not None):
```
NitinAgg, line 42: so this same if was failing in `apply`. How come it would pass here?

NitinAgg (approved):

PR #514, opened 2021-12-03
Feature/re point registry yaml markdowns

bigbitbus:
# Description
Make the registry so the module files (yaml, md) are picked up from the modules/ folder instead of from config/

# Safety checklist

To be tested in CircleCI manually


opta/json_schema.py
```diff
@@ -179,11 +173,14 @@ def _check_module_schemas(write: bool = False) -> None:
                 "module_type": CONFIG_TYPE_ENV
                 if module_registry_dict["environment_module"]
                 else CONFIG_TYPE_SERVICE,
-                "clouds": FOLDER_NAME_TO_CLOUD_LIST[cloud],
+                # "clouds": FOLDER_NAME_TO_CLOUD_LIST[cloud],
+                "clouds": json_schema["opta_metadata"]["clouds"],  # @Quinn ok by you?
```
QuinnNgo, line 177: Discussed in slack but posting here for posterity -- this is great! Totally fine with me!

opta/registry.py
```diff
@@ -99,18 +102,40 @@ def _make_module_docs(vanilla_text: str, module_dict: Dict[Any, Any]) -> str:
...
+    rtn_list = []
+    all_yaml_files = glob.glob(directory + "/**/*.yaml", recursive=True)
+    for a_yaml_path in all_yaml_files:
+        try:
+            module_dict = yaml.load(open(a_yaml_path))
+            if not module_dict:
+                continue
+        except:  # nosec # noqa
+            continue
+        if "clouds" not in module_dict:
```
juandiegopalomino, line 115: when is this the case?

juandiegopalomino (commented):
do we need any changes to make_registry_docs?

juandiegopalomino (approved):
lgtm, but still curious about the "when is this the case" comment I left

bigbitbus:
> do we need any changes to make_registry_docs?

Reference docs are generated correctly.

opta/registry.py
bigbitbus: If there is some other yaml file in the directory (say a test yaml file) then it will be ignored due to this logic since it wont have the clouds key.

PR #515, opened 2021-12-03
Adding support for aws lambda functions

juandiegopalomino:
# Description
see the shiny docs I made for description

# Safety checklist

Tested rigorously by creating both an env for the lambda, having it deployed as a service, deploying w/ and w/o aws-base, with and without public domains


config/registry/aws/modules/lambda-function.md
```diff
@@ -0,0 +1,110 @@
+---
+title: "aws-lambda"
+linkTitle: "aws-lambda"
+date: 2021-12-02
+draft: false
+weight: 1
+description: Creates a lambda deployment via opta.
+---
+
+*NOTE* The lambda module is currently in beta, with only the following functionality. Enhanced features and integrated
```
legalosLOTR, line 10: I think you forgot to type something after integrated

config/registry/aws/modules/lambda-function.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+*NOTE* DO NOT add a subdirectory in the zip file-- make its structure flat unless you know what you're doing (this is
+a common mistake with lambda)
+
+## Accessing your function logs
+As is the standard, all of your function invocation logs are stored in cloudwatch under a new log group. You can
+find the log group name and even a helper shortcut url to them in the `opta output`.
+
+## Expose via Public Domain
```
legalosLOTR, line 64: What domain will this get exposed at? Does api gateway give you a domain by default?

config/registry/aws/modules/lambda-function.yaml
```diff
@@ -0,0 +1,68 @@
...
+    description: |
+      A map of key values to add to the container as environment variables (key is name, value is value).
+      ```yaml
+      env_vars:
+       FLAG: "true"
+      ```
+    default: { }
+  - name: vpc_id
+    user_facing: false
+    description: Opta vpc_id id
```
legalosLOTR, line 48: Does this need to be an opta vpc or could it be any vpc?

modules/lambda_function/lambda-function.md
```diff
@@ -0,0 +1,110 @@
+---
```
legalosLOTR, line 1: Why is this file repeated?

modules/lambda_function/tf_module/api_gateway.tf
```diff
@@ -0,0 +1,57 @@
...
+}
+
+resource "aws_apigatewayv2_integration" "lambda_integration" {
+  count                  = var.expose_via_domain ? 1 : 0
+  api_id                 = aws_apigatewayv2_api.api[0].id
+  integration_type       = "AWS_PROXY"
+  payload_format_version = "2.0"
+  connection_type        = "INTERNET"
+  description            = "Lambda integration"
+  integration_method     = "POST"
```
legalosLOTR, line 22: Are we supporting only POST requests?

legalosLOTR (commented):
This looks great!! Great work!

modules/lambda_function/tf_module/default_package/index.js
```diff
@@ -0,0 +1,8 @@
+exports.handler = async (event) => {
```
legalosLOTR, line 1: What’s the purpose of this file and default.zip?

juandiegopalomino: This is the "hello world" lambda function that is deployed by default if users do not specify a zip file themselves. It's supposed to be used to introduce users to this module giving them a little something they can deploy and see for themselves that it's working well.

modules/lambda_function/tf_module/api_gateway.tf
juandiegopalomino: We're supporting any request method, but the integration with lamba MUST be post-- the data of the original method sent is passed along as part of the event data to the lambda function, as per the documentation

modules/lambda_function/lambda-function.md
juandiegopalomino: Remember the duplication thing Sachin is taking care of? This needs to be here until that is merged

config/registry/aws/modules/lambda-function.md
juandiegopalomino: `Does api gateway give you a domain by default?` yes it does

config/registry/aws/modules/lambda-function.yaml
juandiegopalomino: It needs to be the opta vpc for now because of the way we find the subnets for it. Not opposed to finding a way to make it work with non-opta vpcs in the future, but that's beyond anything we talked about for v1 (also note that this input is hidden from the user--it's automatically filled out if opta detects that an aws-base module is present)

config/registry/aws/modules/lambda-function.md
```diff
@@ -0,0 +1,110 @@
...
+    region: us-east-1
+    account_id: XXXXXXXXX
+modules:
+  - type: lambda-function
+    expose_via_domain: true
+```
+
+## Adding your Own Function
+Opta supports the [zip file form](https://docs.aws.amazon.com/lambda/latest/dg/configuration-function-zip.html) of
+lambda deployments. For this deployment, simply zip up your code into a zip file and pass its location in via the
```
legalosLOTR, line 38: So I like the idea of a default deployment lambda but it shouldn't be baked into opta code. We should provide a link to an example zip file in our docs and tell the user "if you wanna try this out just download this zip and put a link to it in your opta.yml".
And then, we should make the `filename` field required.

config/registry/aws/modules/lambda-function.md
```diff
@@ -0,0 +1,110 @@
...
+
+*NOTE* DO NOT add a subdirectory in the zip file-- make its structure flat unless you know what you're doing (this is
+a common mistake with lambda)
+
+## Accessing your function logs
+As is the standard, all of your function invocation logs are stored in cloudwatch under a new log group. You can
+find the log group name and even a helper shortcut url to them in the `opta output`.
+
+## Expose via Public Domain
+You can have your lambda be automatically exposed to the world via a public ui by setting the `expose_via_domain` field
```
legalosLOTR, line 65: Can you add a line here like "you can get the default domain by running opta output"?

modules/lambda_function/lambda-function.md
```diff
@@ -20,7 +20,8 @@ much more setup. Currently, it also handles:
 * Allowing the user to pass in IAM policies to give to the lambda
 * Setting up security group and network location if you have the `aws-base` module in your environment.
 
-tl;dr -- this will deploy a hello world lambda
+tl;dr -- Download [this zip](https://gist.github.com/juandiegopalomino/7400f836107459f3099c02e58d2d6897/raw/5c0b4cbbbf1d8a470f16c09d53e6ed68e59e06bd/baloney.zip),
```
legalosLOTR, line 23: Slightly better way to write this: "Simple way to try out this module: Just download this example zip file, put it in the same directory as your opta.yml and run opta apply".

legalosLOTR: Also, it's better to put the zip file in our docs static dir (https://github.com/run-x/opta-docs/tree/main/static) - instead of a personal gist.

modules/lambda_function/lambda_function.py
```diff
@@ -32,8 +32,8 @@ def process(self, module_idx: int) -> None:
                 else f"module.{aws_base_module.name}"
             )
             self.module.data["vpc_id"] = f"${{{{{module_source}.vpc_id}}}}"
-        file_path: Optional[str] = self.module.data.get("filename")
-        if file_path is not None and not file_path.startswith("/"):
+        file_path: str = self.module.data.get("filename")
+        if file_path.startswith("/"):
```
legalosLOTR, line 36: What's going on here?

juandiegopalomino: Handling relative paths

modules/lambda_function/lambda-function.md
juandiegopalomino: We got a chicken and the egg scenario here. Once the lambda is deployed and docs first updated, we can make an update to point to a zip file in our docs.

config/registry/aws/modules/lambda-function.md
```diff
@@ -0,0 +1,115 @@
+---
+title: "aws-lambda"
+linkTitle: "aws-lambda"
+date: 2021-12-02
+draft: false
+weight: 1
+description: Creates a lambda deployment via opta.
+---
+
+*NOTE* The lambda module is currently in beta, with only the following functionality. Enhanced features and integrated
```
legalosLOTR, line 10: Looks like the typo in this line was brought back in the latest commits.

legalosLOTR (approved):

PR #516, opened 2021-12-06
Apply env if it is not found by service

juandiegopalomino:
# Description
Apply env if it is not found by service

# Safety checklist

Tested locally.

### DIFF
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -362,7 +362,7 @@ def _verify_parent_layer(layer: Layer) -> None:
         click.confirm(
             f"Failed to get the Environment state {e.args[0]} "
             "Usually, this means that the Environment mentioned in configuration file does not exist. \n"
-            "Would you like to create your environment?",
+            f"Would you like to create your environment using {layer.parent.path}?",
             abort=True,
         )
         _apply(
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -1,5 +1,7 @@
 import datetime
 import os
+import shutil
+import tempfile
 from pathlib import Path
 from subprocess import CalledProcessError  # nosec
 from threading import Thread
@@ -13,6 +15,7 @@ from botocore.config import Config
 from botocore.exceptions import ClientError
 
 from opta.amplitude import amplitude_client
+from opta.cleanup_files import cleanup_files
 from opta.commands.local_flag import _clean_tf_folder, _handle_local_flag
 from opta.constants import DEV_VERSION, TF_PLAN_PATH, VERSION
 from opta.core.aws import AWS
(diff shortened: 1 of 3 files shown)


opta/commands/apply.py
```diff
@@ -358,8 +361,26 @@ def _verify_parent_layer(layer: Layer) -> None:
...
             )
     except MissingState as e:
-        raise MissingState(
-            f"Failed to get the Environment state {e.args[0]}"
-            "Usually, this means that the Environment mentioned in configuration file does not exist."
-            "You can read more about this in our getting started guide: https://docs.opta.dev/getting-started/"
+        click.confirm(
+            f"Failed to get the Environment state {e.args[0]} "
+            "Usually, this means that the Environment mentioned in configuration file does not exist. \n"
+            "Would you like to create your environment?",
```
bigbitbus, line 367: It would be nice to also tell the user which environment file will be used (this info should be in the service yaml file). For example the question  could be

"Would you like to create your environment using /home/foo/optafiles/fooenv.yaml?"

bigbitbus (approved):
Looks good to me; there is a minor comment in the review if you want to consider that.

opta/commands/apply.py
```diff
@@ -358,8 +361,26 @@ def _verify_parent_layer(layer: Layer) -> None:
...
+            abort=True,
         )
+        t = tempfile.mkdtemp()
+        try:
+            config_path = os.path.join(t, "opta.yml")
+            with open(config_path, "w") as f:
+                f.write(layer.parent.original_spec)
```
legalosLOTR, line 374: Why can't we just provide the path to the env file to `_apply`? Instead of creating this temp copy.

opta/commands/apply.py
```diff
@@ -358,8 +359,19 @@ def _verify_parent_layer(layer: Layer) -> None:
...
-            "Usually, this means that the Environment mentioned in configuration file does not exist."
-            "You can read more about this in our getting started guide: https://docs.opta.dev/getting-started/"
+        click.confirm(
+            f"Failed to get the Environment state {e.args[0]} "
+            "Usually, this means that the Environment mentioned in configuration file does not exist. \n"
+            "Would you like to create your environment?",
+            abort=True,
         )
+        _apply(
+            layer.parent.path,
```
NitinAgg, line 369: are there any relative path issues here?

opta/layer.py
```diff
@@ -222,7 +222,7 @@ def load_from_yaml(
             raise UserErrors(f"File {config} not found")
 
         conf["original_spec"] = config_string
-        conf["path"] = config
+        conf["path"] = config_path
```
legalosLOTR, line 225: This sounds like a series issue.. was this broken before?

legalosLOTR (approved):

opta/layer.py
juandiegopalomino: great question, it could have been, but this field was rarely ever used in the regular layer, and never used by the parent (where the git redirect issue was present)

PR #519, opened 2021-12-07
Separate the Unlock and Helm Command.

nsarupr:
# Description
Separate the Terraform Unlock and Helm Release command.

# Safety checklist

Tested Manually by Creating the Environment in AWS and GCP. Tested Force Unlock before and after k8s-cluster creation,

### DIFF
--- a/opta/commands/force_unlock.py
+++ b/opta/commands/force_unlock.py
@@ -28,23 +28,66 @@ def force_unlock(config: str, env: Optional[str]) -> None:
     amplitude_client.send_event(amplitude_client.FORCE_UNLOCK_EVENT)
     layer = Layer.load_from_yaml(config, env)
     layer.verify_cloud_credentials()
+    modules = Terraform.get_existing_modules(layer)
+    layer.modules = [x for x in layer.modules if x.name in modules]
     gen_all(layer)
 
-    Terraform.init(layer=layer)
+    if Terraform.download_state(layer):
+        tf_lock_exists, _ = Terraform.tf_lock_details(layer)
+        if tf_lock_exists:
+            Terraform.init(layer=layer)
+            click.confirm(
+                "This will remove the lock on the remote state."
+                "\nPlease make sure that no other instance of opta command is running on this file."
+                "\nDo you still want to proceed?",
+                abort=True,
+            )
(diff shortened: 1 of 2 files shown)


juandiegopalomino (approved):
lgtm

opta/commands/force_unlock.py
```diff
@@ -28,23 +28,66 @@ def force_unlock(config: str, env: Optional[str]) -> None:
...
+            tf_flags.append("-force")
+            Terraform.force_unlock(layer, *tf_flags)
 
-    click.confirm(
-        "This will remove the lock on the remote state."
-        "\n\tPlease make sure that no other instance of opta command is running on this file."
-        "\n\tDo you still want to proceed?",
-        abort=True,
-    )
+        if "k8scluster" in modules:
```
NitinAgg, line 48: this won't work on a service file right?

NitinAgg, line 48: this won't work on a service file right?

opta/commands/force_unlock.py
```diff
@@ -28,23 +28,66 @@ def force_unlock(config: str, env: Optional[str]) -> None:
...
+            configure_kubectl(layer)
```
NitinAgg, line 49: I think there is a case where k8scluster is in modules but has not successfully run yet. so this would fail..

NitinAgg: Maybe a better check would be if tf outputs have the k8s cluster url?

opta/commands/force_unlock.py
nsarupr: Okay, so I didn't check it for the service files. Let me check it there as well.

opta/commands/force_unlock.py
nsarupr: Here we are checking if the whole of the k8scluster module is created successfully, and that only appears once a single terraform apply succeeds.

nsarupr:
Thanks for the Catch @NitinAgg . I have added the checks for Service Configuration. Please check. 
@juandiegopalomino can you review the changes as well, in case I missed something.

NitinAgg (approved):

PR #520, opened 2021-12-07
Jd/checker script

juandiegopalomino:
# Description
This adds a dictionary in constants where the key is a semver version, and the value is an upgrade warning to that version. Now if we ever need to give a heads up for a future breaking change, all we gotta do is add entries to that dict.

# Safety checklist

Made new tests and tried running through cli

### DIFF
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -15,7 +15,7 @@ from botocore.exceptions import ClientError
 from opta.amplitude import amplitude_client
 from opta.cleanup_files import cleanup_files
 from opta.commands.local_flag import _clean_tf_folder, _handle_local_flag
-from opta.constants import DEV_VERSION, TF_PLAN_PATH, VERSION
+from opta.constants import DEV_VERSION, TF_PLAN_PATH, UPGRADE_WARNINGS, VERSION
 from opta.core.aws import AWS
 from opta.core.azure import Azure
 from opta.core.cloud_client import CloudClient
@@ -194,26 +194,17 @@ def _apply(
     else:
         raise Exception(f"Cannot handle upload config for cloud {layer.cloud}")
 
-    previous_config: Optional[StructuredConfig] = cloud_client.get_remote_config()
-    if previous_config is not None and "opta_version" in previous_config:
-        old_opta_version = previous_config["opta_version"]
-        if (
-            old_opta_version not in [DEV_VERSION, ""]
-            and VERSION not in [DEV_VERSION, ""]
-            and semver.VersionInfo.parse(VERSION.strip("v")).compare(
-                old_opta_version.strip("v")
-            )
-            < 0
-        ):
-            raise UserErrors(
-                f"ou're trying to run an older version of opta (last run was at {previous_config['date']} with version {old_opta_version}). Please update to the latest version and try again!"
-            )
+    existing_config: Optional[StructuredConfig] = cloud_client.get_remote_config()
+    if existing_config is not None and "opta_version" in existing_config:
+        old_semver_string = existing_config["opta_version"].strip("v")
+        current_semver_string = VERSION.strip("v")
+        _verify_semver(old_semver_string, current_semver_string)
 
     try:
         existing_modules: Set[str] = set()
         first_loop = True
         for module_idx, current_modules, total_block_count in gen(
-            layer, previous_config, image_tag, image_digest, test, True
+            layer, existing_config, image_tag, image_digest, test, True
         ):
             if first_loop:
                 # This is set during the first iteration, since the tf file must exist.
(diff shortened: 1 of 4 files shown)


opta/commands/apply.py
```diff
@@ -194,26 +195,17 @@ def _apply(
...
-            and semver.VersionInfo.parse(VERSION.strip("v")).compare(
-                old_opta_version.strip("v")
-            )
-            < 0
-        ):
-            raise UserErrors(
-                f"ou're trying to run an older version of opta (last run was at {previous_config['date']} with version {old_opta_version}). Please update to the latest version and try again!"
-            )
+    existing_config: Optional[StructuredConfig] = cloud_client.get_remote_config()
+    if existing_config is not None and "opta_version" in existing_config:
```
NitinAgg, line 199: Let's move this whole block into _verify_semver?

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
         )
 
 
+def _verify_semver(old_semver_string: str, current_semver_string: str) -> None:
+    if old_semver_string not in [DEV_VERSION, ""] and current_semver_string not in [
```
NitinAgg, line 330: nit: a slightly cleaner way is if you check `if old_semver_string in [DEV_VERSION, ""]...; return` Then you won't have to indent the whole block below this

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
...
+        DEV_VERSION,
+        "",
+    ]:
+        old_semver = semver.VersionInfo.parse(old_semver_string)
+        current_semver = semver.VersionInfo.parse(current_semver_string)
+        if old_semver > current_semver:
+            raise UserErrors(
```
NitinAgg, line 337: We already have this test somewhere right? maybe we should remove that as well

opta/constants.py
```diff
@@ -24,6 +25,12 @@
 DEV_VERSION = "dev"
 
 SESSION_ID = int(time.time() * 1000)
+UPGRADE_WARNINGS: Dict[str, str] = {
+    "0.21.0": "Performing updates on AWS environments to the network load balancer which will cause a new one to be "
```
NitinAgg, line 29:
```suggestion
    "0.21.0": "If you are upgrading an AWS environment to a newer version, this upgrade will cause a 5 min downtime for any public traffic. This will update the network load balancer which will cause a new one to spin up."
```

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
         )
 
 
+def _verify_semver(old_semver_string: str, current_semver_string: str) -> None:
+    if old_semver_string not in [DEV_VERSION, ""] and current_semver_string not in [
+        DEV_VERSION,
+        "",
+    ]:
```
rabbitfang, line 333:
```suggestion
    if old_semver_string in [DEV_VERSION, ""] or current_semver_string in [
        DEV_VERSION,
        "",
    ]:
        return
```

Then un-indent the rest of the lines in this function by one level.

opta/constants.py
```diff
@@ -24,6 +25,12 @@
 DEV_VERSION = "dev"
 
 SESSION_ID = int(time.time() * 1000)
+UPGRADE_WARNINGS: Dict[str, str] = {
```
rabbitfang, line 28: Can you document the format of this constant? Is the key the last version where the message is *not* printed? or is it the first version where we print the message?

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
...
+            ],
+            key=lambda x: semver.VersionInfo.parse(x[0]),
+        )
+        for current_upgrade_warning in current_upgrade_warnings:
+            logger.info(
+                f"{fg('magenta')}WARNING{attr(0)}: Detecting an opta upgrade to or past version {current_upgrade_warning[0]}. "
+                f"Got the following warning: {current_upgrade_warning[1]}"
+            )
+        if len(current_upgrade_warnings) > 0:
+            click.confirm(
```
rabbitfang, line 356: Do we want to have a way of bypassing this confirmation dialog via CLI arguments?

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
...
+            raise UserErrors(
+                f"You're trying to run an older version of opta (last run with version {old_semver}). "
+                "Please update to the latest version and try again!"
+            )
+
+        current_upgrade_warnings = sorted(
+            [
+                (k, v)
+                for k, v in UPGRADE_WARNINGS.items()
+                if current_semver.compare(k) >= 0 and old_semver.compare(k) < 0
```
rabbitfang, line 346:
```suggestion
                if current_semver >= k and old_semver < k
```

Strings will be implicitly converted to `VersionInfo` objects by the comparison operators.

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
...
+            ],
+            key=lambda x: semver.VersionInfo.parse(x[0]),
+        )
+        for current_upgrade_warning in current_upgrade_warnings:
+            logger.info(
+                f"{fg('magenta')}WARNING{attr(0)}: Detecting an opta upgrade to or past version {current_upgrade_warning[0]}. "
+                f"Got the following warning: {current_upgrade_warning[1]}"
+            )
+        if len(current_upgrade_warnings) > 0:
```
rabbitfang, line 355:
```suggestion

        if current_upgrade_warnings:
```

opta/commands/apply.py
```diff
@@ -334,6 +326,39 @@ def _apply(
...
+                for k, v in UPGRADE_WARNINGS.items()
+                if current_semver.compare(k) >= 0 and old_semver.compare(k) < 0
+            ],
+            key=lambda x: semver.VersionInfo.parse(x[0]),
+        )
+        for current_upgrade_warning in current_upgrade_warnings:
+            logger.info(
+                f"{fg('magenta')}WARNING{attr(0)}: Detecting an opta upgrade to or past version {current_upgrade_warning[0]}. "
+                f"Got the following warning: {current_upgrade_warning[1]}"
+            )
```
rabbitfang, line 354: Why are we using the INFO logging level but prefixing the text with "WARNING"? This should probably be `logger.warning`

rabbitfang (changes requested):

opta/commands/apply.py
juandiegopalomino: I don't think so for now? It seems superfluous

opta/commands/apply.py
juandiegopalomino: I don't think we do

opta/commands/apply.py
juandiegopalomino: I like to be more explicit

opta/commands/apply.py
juandiegopalomino: It's a warning to the user, but not to the system

opta/commands/apply.py
juandiegopalomino: done

opta/constants.py
juandiegopalomino: done

opta/commands/apply.py
juandiegopalomino: done

opta/constants.py
juandiegopalomino: updated with better text

opta/commands/apply.py
juandiegopalomino: done

opta/commands/apply.py
juandiegopalomino: addressed

NitinAgg (approved):

rabbitfang (approved):

PR #521, opened 2021-12-07
Revived custom terraform module

juandiegopalomino:
# Description
With fancy docs

# Safety checklist

See demo in docs

### DIFF
--- a/config/registry/schemas/modules/custom-terraform.json
+++ b/config/registry/schemas/modules/custom-terraform.json
@@ -1,0 +1,34 @@
+{
+  "$id": "https://app.runx.dev/modules/custom-terraform",
+  "description": "To bring in your own custom terraform module",
+  "type": "object",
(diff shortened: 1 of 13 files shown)


NitinAgg (approved):

config/registry/schemas/modules/custom-terraform.json
```diff
@@ -0,0 +1,34 @@
...
+    "clouds": [
+      "aws",
+      "gcp"
+    ]
+  },
+  "required": [
+    "path_to_module",
+    "type"
+  ]
+}
```
bigbitbus, line 34: I see I have not removed the config/schemas/modules directory yet, doing that in another PR. 

This file need not be put into  config/registry/schemas/modules/, the single copy in modules/custom_terraform is adequate.

bigbitbus (approved):

PR #522, opened 2021-12-08
Update JSON Schema script and fix typos

QuinnNgo:
# Description
This PR updates the JSON Schema script to add some quality-of-life things required by the frontend including:
- a displayName for each module
- extra checks for fields required in json schemas
- an assortment of fixed typos (e.g. `"type": "list"` -> `"type": array`)

The only changed file of substance is the `json_schema.py` script

# Safety checklist


opta/json_schema.py
```diff
@@ -116,18 +116,21 @@ def _check_opta_config_schemas(write: bool = False) -> None:
...
 
 def _check_module_schemas(write: bool = False) -> None:
-    for cloud in CLOUD_FOLDER_NAMES + ["common"]:
+    for cloud in ["aws", "azurerm", "google", "local"]:
         if cloud != "common":
             index_yaml = yaml.load(open(join(registry_path, cloud, "index.yaml")))
 
-        module_info = _get_all_module_info(modules_path, cloud)
+        module_info = _get_all_module_info(
+            modules_path, CLOUD_NAME_TO_JSON_SCHEMA_NAME[cloud]
```
QuinnNgo, line 124: @bigbitbus Found this tiny bug where the `azure` and `gcp` schemas weren't getting picked up because of our name flipping between `azurerm`/`azure` and `google`/`gcp`

bigbitbus (approved):
Looks good to me, I havent gone through all the json schema changes but the py code change looks good.

PR #526, opened 2021-12-08
Opta Local Tests added to Circle CI

nsarupr:
# Description
Add Circle CI tests for Opta Local

# Safety checklist

YOUR_ANSWER


juandiegopalomino:
@nsarupr rebase to ignore the CI/python failure (that was my bad). Also, could you actually run your new workflow to show it's all good?

NitinAgg (approved):

PR #529, opened 2021-12-09
Removed redundant credential linking code

bigbitbus:
# Description
Removing redundant linker credential code from the 3 k8s_services files, as well as some unit tests that are no longer needed given the removal.

# Safety checklist

Checking in CCI


juandiegopalomino (approved):
lgtm

PR #533, opened 2021-12-12
Adding a wait for in opta base cause it keeps race conditioning

juandiegopalomino:
# Description
Like right here: https://app.circleci.com/pipelines/github/run-x/opta/305/workflows/7f590e3d-1b1c-40de-a7ab-c88c52a44d25/jobs/689

So now opta base waits 30 seconds to make sure all the things it depends on are up and running smoothly

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/aws_k8s_base/tf_module/opta_base.tf
+++ b/modules/aws_k8s_base/tf_module/opta_base.tf
@@ -10,6 +10,17 @@ resource "helm_release" "opta_base" {
     })
   ]
   depends_on = [
-    helm_release.cert_manager
+    time_sleep.wait_a_bit
   ]
+}
+
+resource "time_sleep" "wait_a_bit" {
+  depends_on = [
+    helm_release.cert_manager,
+    helm_release.autoscaler,
+    helm_release.load_balancer,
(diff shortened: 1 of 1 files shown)


modules/aws_k8s_base/tf_module/opta_base.tf
```diff
@@ -10,6 +10,17 @@ resource "helm_release" "opta_base" {
     })
   ]
   depends_on = [
-    helm_release.cert_manager
+    time_sleep.wait_a_bit
   ]
+}
+
+resource "time_sleep" "wait_a_bit" {
```
NitinAgg, line 17: Let's just try by putting all these dependencies in the depends on above. We have too many sleeps in opta now. That will slow opta down a lot

juandiegopalomino: We use this exactly once and that's for gcp. Even if we did, the apply sometimes fails on the cert manager check because waiting for it to be deployed is not enough-- it needs a few more seconds to be operational

NitinAgg (approved):

PR #535, opened 2021-12-13
Adding option to increase log retention periods

juandiegopalomino:
# Description
As requested

# Safety checklist

n/a


NitinAgg (approved):

bigbitbus (commented):
Just want to confirm (you may have seen this in testing)
  * these numbers ("defaults") are exactly equal to what the current default is set to? 
  * this change does not result in re-creation of resources in the vpc

Want to make sure this change does not change the status quo of existing deployments.

juandiegopalomino:
@bigbitbus yup, log retention periods can easily be changed

PR #542, opened 2021-12-14
EKS Cloudfront.

nsarupr:
# Description
Add Cloudfront Distribution on top of EKS Load Balancer.

# Safety checklist

Tested manually by creating and destroying the Cloudfront distribution.


modules/cloudfront_distribution/tf_module/main.tf
```diff
@@ -7,9 +8,68 @@ data "aws_s3_bucket" "logging_bucket" {
   bucket = var.s3_log_bucket_name
 }
 
+locals {
+  lb_distribution_count = var.eks_load_balancer_enabled == true ? 1 : 0
+  s3_distribution_count = var.s3_load_balancer_enabled == true ? 1 : 0
+}
+
+resource "aws_cloudfront_distribution" "lb_distribution" {
```
juandiegopalomino, line 16: @NitinAgg thoughts on having a separate resource? I know I dug my own grave by first calling it `s3_distribution`, but we ought to just have a single resource so that we can easily update to support going to both s3 AND eks depending on path.

NitinAgg: Agree. Let's try to fit this in the existing module. We can change the resource name. No one is using this in prod yet

modules/cloudfront_distribution/tf_module/main.tf
```diff
@@ -7,9 +8,68 @@ data "aws_s3_bucket" "logging_bucket" {
...
+    for_each = var.s3_log_bucket_name == null ? [] : [1]
+    content {
+      include_cookies = true
+      bucket          = data.aws_s3_bucket.logging_bucket[0].bucket_domain_name
+      prefix          = "cloudfront/${var.layer_name}/${var.module_name}"
+    }
+  }
+
+  default_cache_behavior {
+    allowed_methods        = ["GET", "HEAD", "OPTIONS"]
```
NitinAgg, line 43: what's the difference b/w allowed and cached? and yes, it's good idea to expose the cached methods to users but as optional. and you can use this list as default if the user doesn't specify it

nsarupr: `allowed_methods`: Controls which HTTP methods CloudFront processes and forwards to your Amazon S3 bucket or your custom origin.
`cached_methods`: Controls whether CloudFront caches the response to requests using the specified HTTP methods.

Can give access to the users for both of them. Or should we do that in the next phase?

modules/cloudfront_distribution/tf_module/main.tf
nsarupr: @NitinAgg @juandiegopalomino @bigbitbus have a look at these changes.
For references, you can also have a look at these Cloudfront Distributions (Runx-Tmp AWS Account) as well:
* E2TICDHTZE41Y0
* E11KTM0ZEFNT3A

modules/cloudfront_distribution/tf_module/main.tf
```diff
@@ -31,61 +30,99 @@ resource "aws_cloudfront_distribution" "s3_distribution" {
...
+
+        cookies {
+          forward = "none"
+        }
+      }
+    }
+  }
 
-  price_class = var.price_class
+  dynamic "default_cache_behavior" {
```
juandiegopalomino, line 90: duplicate block, with only difference seeming to be target_origin_id

nsarupr: merged both the blocks into one.

modules/cloudfront_distribution/cloudfront-distribution.yaml
```diff
@@ -62,6 +74,16 @@ inputs:
...
     default: []
+  - name: allowed_methods
+    user_facing: true
+    validator: list(str(), required=False)
+    description: HTTP methods CloudFront processes and forwards to your Amazon S3 bucket or your custom origin.
+    default: ["GET", "HEAD", "OPTIONS"]
+  - name: cached_methods
+    user_facing: true
+    validator: list(str(), required=False)
+    description: CloudFront caches the response to requests using the specified HTTP methods.
```
bigbitbus, line 85: CloudFront caches the response to the specified HTTP method requests.

bigbitbus (approved):
Comments mostly around documentation, users will need a tutorial.

modules/cloudfront_distribution/cloudfront-distribution.yaml
```diff
@@ -62,6 +74,16 @@ inputs:
     validator: list(str(), required=False)
     description: The linked s3 buckets to attach to your cloudfront distribution (currently only supports one).
     default: []
+  - name: allowed_methods
+    user_facing: true
+    validator: list(str(), required=False)
+    description: HTTP methods CloudFront processes and forwards to your Amazon S3 bucket or your custom origin.
```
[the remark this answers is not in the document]
nsarupr: We can write a blog post for the same.

NitinAgg (approved):

PR #546, opened 2021-12-16
Feature/dynamic module load

bigbitbus:
# Description
Load Opta modules dynamically instead of loading all of them in the layer.py file.

# Safety checklist

Manual testing


opta/registry.py
```diff
@@ -50,7 +51,7 @@ def make_registry_dict() -> Dict[Any, Any]:
             alt_cloudname = cloud
         cloud_dict["modules"] = {**_make_module_registry_dict(module_path, alt_cloudname)}
         registry_dict[cloud] = cloud_dict
-
+    open("/tmp/registry.json", "w").write(json.dumps(registry_dict))  # nosec
```
rabbitfang, line 54: Was this meant to be left in?

opta/layer.py
```diff
@@ -75,6 +35,55 @@
 from opta.utils import deep_merge, hydrate, logger, yaml
 from opta.utils.dependencies import validate_installed_path_executables
 
+PROCESSOR_DICT: Dict[str, str] = {
```
rabbitfang, line 38: Can you please add a unit test that checks that all of the processors in this dict exist? Something like:
```python
for processor in PROCESSOR_DICT.values():
    pymodule_path = generate_pymodule_path(processor)
    importlib.import_module(pymodule_path)
```

rabbitfang (changes requested):

opta/registry.py
bigbitbus: no, it was debug info, removed.

opta/layer.py
bigbitbus: Added a test to check if each module in the dictionary is actually in the directories.

bigbitbus:
@rabbitfang addressed the comments in the review. @juandiegopalomino as discussed when we use this in the new module api we will try to automatically infer the mapping instead of the dictionary. For now with @rabbitfang suggestion I added a test to fail if the module code does not exist at the expected location in the module directory.

tests/test_layer.py
```diff
@@ -322,3 +376,11 @@ def test_child(self, mocker: MockFixture):
...
             ]
         )
+
+    def test_check_module_dirs_exist(self):
+        from opta.layer import PROCESSOR_DICT, get_processor_class
+
+        for opta_module, processor in PROCESSOR_DICT.items():
+            module_class = get_processor_class(opta_module)
+            assert module_class.__name__ == processor
+        print("done")
```
rabbitfang, line 386:
```suggestion
```

rabbitfang (approved):
Everything else LGTM

bigbitbus:
@rabbitfang added your commit.

PR #549, opened 2021-12-20
Remove Max Version limitation from Opta

nsarupr:
# Description
Remove Max Version limitation from Opta

# Safety checklist

N/A


opta/constants.py
```diff
@@ -42,7 +42,7 @@
 OPTA_DISABLE_REPORTING = "OPTA_DISABLE_REPORTING"
 TF_FILE_PATH = "main.tf.json"
 TF_PLAN_PATH = "tf.plan"
-MAX_TERRAFORM_VERSION = "1.1.0"
+# MAX_TERRAFORM_VERSION = "1.1.0"
```
RemyDeWolf, line 45: Should we proactively not support v2 (would be the next major version)?

opta/core/terraform.py
```diff
@@ -117,10 +118,10 @@ def validate_version(cls) -> None:
                 f"Invalid terraform version {current_version} -- must be at least {MIN_TERRAFORM_VERSION}"
             )
 
-        if current_parsed >= version.parse(MAX_TERRAFORM_VERSION):
-            raise UserErrors(
-                f"Invalid terraform version {current_version} -- must be less than {MAX_TERRAFORM_VERSION}"
-            )
+        # if current_parsed >= version.parse(MAX_TERRAFORM_VERSION):
```
RemyDeWolf, line 121: if we not use some code, we can remove it to stay lean.
There is always github history to find previous versions.

opta/constants.py
nsarupr: A good point made...we should do that... let me make this `2.0.0`. That should suffice I guess?

RemyDeWolf: Yes

juandiegopalomino:
what remy said. Also, never leave commented out code in prs

RemyDeWolf (approved):

PR #551, opened 2021-12-21
AWS MySQL Module

nsarupr:
# Description
Adding a new Module for AWS: **MySQL**
Note: As of now, Performance Insights have been disabled. 

# Safety checklist

Tested manually by creating

### DIFF
--- a/modules/aws_mysql/tf_module/main.tf
+++ b/modules/aws_mysql/tf_module/main.tf
@@ -1,4 +1,4 @@
-data "aws_security_group" "mysql_security_group" {
+data "aws_security_group" "security_group" {
   name = "opta-${var.env_name}-db-sg"
 }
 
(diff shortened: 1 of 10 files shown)


modules/aws_mysql/tf_module/main.tf
```diff
@@ -0,0 +1,49 @@
...
+data "aws_kms_key" "main" {
+  key_id = "alias/opta-${var.env_name}"
+}
+
+resource "random_password" "mysql_password" {
+  length  = 20
+  special = false
+}
+
+resource "random_string" "db_name_hash" {
```
RemyDeWolf, line 14: would that make sense to make this name customizable by the user?

I wonder how it will work if the user would like to create 2 mysql DB and what would be the way for them to differentiate them? 

Also if they want to monitor the DB (for example in datadog or cloudwatch) would that be easy for them to differentiate their DB?

I might be missing something here, may be the layer_name would contain some relevant info?

nsarupr: That's just a random 4 letter hash appended to the end of the Actual Identifier.
Example: 
Template: `"opta-${var.layer_name}-${var.module_name}-${random_string.db_name_hash.result}"`
Sample: `opta-mysql-ci-mysql-jl72`

RemyDeWolf: so module name can be used to identify their DB?

for example:
```
  - name: purchase
     type: mysql
  - name: inventory
     type: mysql
```

nsarupr: > so module name can be used to identify their DB?
> 
> for example:
> 
> ```
>   - name: purchase
>     type: mysql
>   - name: inventory
>     type: mysql
> ```

Yes, that's how we have implemented this.
Although I'll check this and confirm.

bigbitbus (commented):
Have we got the linker code for the secrets included here for mysql, for example this file

https://github.com/run-x/opta/blob/aa2cbd4e9848be1562dcca194a6340498789ace4/modules/aws_k8s_service/aws_k8s_service.py#L78

modules/aws_mysql/tf_module/main.tf
nsarupr: Works.

nsarupr:
> Have we got the linker code for the secrets included here for mysql, for example this file
> 
> https://github.com/run-x/opta/blob/aa2cbd4e9848be1562dcca194a6340498789ace4/modules/aws_k8s_service/aws_k8s_service.py#L78

This part is yet to be taken into account. Currently working on the Implementation for the Database Itself. Will work on the linker later.

juandiegopalomino (approved):
lgtm

PR #552, opened 2021-12-21
Added a generic try-except block to catch any datadog logging errors …

bigbitbus:
…while running nice subprocess

# Description

# Safety checklist

Manual test

### DIFF
--- a/opta/nice_subprocess.py
+++ b/opta/nice_subprocess.py
@@ -24,19 +24,24 @@ nice_logger.propagate = False
 
 
 def log_to_datadog(msg: str, severity: str) -> None:
-    msg = ansi_scrub(msg)
-    if hasattr(sys, "_called_from_test") or VERSION == DEV_VERSION or not VERSION:
-        if os.environ.get("OPTA_DEBUG", "") == "DATADOG_LOCAL":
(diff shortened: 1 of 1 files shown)


NitinAgg (approved):

juandiegopalomino (approved):

opta/nice_subprocess.py
```diff
@@ -24,19 +24,24 @@
...
-    else:
-        nice_logger.info(msg)
+        if severity == "ERROR":
+            nice_logger.error(msg)
+        else:
+            nice_logger.info(msg)
+    except:  # nosec   # noqa: E722
+        # Logging error to datadog SaaS, happens,
+        # not the end of world, silently fail
+        pass
```
RemyDeWolf, line 44: I wonder if we would want to see this error _only_ if `OPTA_DEBUG` is set.

bigbitbus: > I wonder if we would want to see this error _only_ if `OPTA_DEBUG` is set.

The underlying issue that casued this hotfix was that datadog endpoint wasnt reachable by the user who reported the problem; in all debug runs we internally usually don't send logs to datadog.

PR #553, opened 2021-12-21
Rename the default config file from opta.yml to opta.yaml

RemyDeWolf:
# Description
Rename the default config file to use the `yaml` extension.
This is consistent with the [kubernetes official documentation](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) and [examples](https://github.com/kubernetes/kubernetes/search?q=filename%3A*.yaml) where yaml is preferred.

# Safety checklist
    If  a user has defined the default opta.yml file, they will see this warning message:
```
WARNING: Could not find file opta.yaml, but loaded opta.yml.
```

Ran `opta apply` locally with `opta.yml` and `opta.yaml` to test the behavior.


NitinAgg (commented):
This is a backwards incompatible change but should be fine. cc @legalosLOTR

RemyDeWolf:
@NitinAgg some tests were failing (coming from https://github.com/run-x/test-service)
I added some changes to support both opta.yml and opta.yaml.
A warning will be showed when opta.yml is used as the default file, see the PR description for more info.

opta/utils/__init__.py
```diff
@@ -202,8 +202,30 @@ def exp_backoff(num_tries: int = 3) -> Generator:
...
+    if ext.lower() == ".yaml":
+        return pre + ".yml", True
+    elif ext.lower() == ".yml":
+        return pre + ".yaml", True
+    return pre + ext, False
+
+
 def check_opta_file_exists(config_path: str) -> str:
     if not os.path.exists(config_path):
+        # try alternate y(a)ml extension
```
NitinAgg, line 221: Missed this earlier. But this is brilliant!!

NitinAgg (approved):

PR #555, opened 2021-12-22
Support alternate y(a)ml extension for environment path

RemyDeWolf:
# Description
Following #553, the next step is to rename our opta files in all our repos from `.yml` to `.yaml`.
The problem we will face is that there are many files that reference environment files in other repos. When renaming them we would break their links.

ex:
1. we rename https://github.com/run-x/runx-infra/blob/main/staging/opta.yml to opta.yaml
2. it will break the test-service which references `git@github.com:run-x/runx-infra.git//staging/opta.yml?ref=main"` ([here]
3. this will break the CI check: CI / deploy_test_service

If some users have multiple github repos to store their opta files, they might run into a similar issue.

This PR will address this issue by supporting both yaml/yml extension for the environment path.

ex:
1. we rename https://github.com/run-x/runx-infra/blob/main/staging/opta.yml to opta.yaml
2. the test-service is not yet changed and still references `git@github.com:run-x/runx-infra.git//staging/opta.yml?ref=main"` ([here]
3. CI / deploy_test_service: will work because it will try `git@github.com:run-x/runx-infra.git//staging/opta.yaml` if `git@github.com:run-x/runx-infra.git//staging/opta.yml` is not found.

# Safety checklist

This change could be tested by adding a few unit tests included in this PR.

### DIFF
--- a/opta/layer.py
+++ b/opta/layer.py
@@ -7,7 +7,6 @@ import re
 import shutil
 import tempfile
 from datetime import datetime
-from os import path
 from pathlib import Path
 from types import SimpleNamespace
 from typing import Any, Dict, FrozenSet, Iterable, List, Optional, Set, Tuple, TypedDict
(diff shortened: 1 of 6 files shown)


juandiegopalomino (approved):
lgtm

PR #556, opened 2021-12-23
Hotfix/build imports issue fix

bigbitbus:
# Description
Explicitly import modules when building opta binary since dynamic imports are not detected by pyinstaller.

# Safety checklist

Manually


opta.spec
```diff
@@ -4,7 +4,37 @@ from PyInstaller.utils.hooks import collect_data_files
 
 block_cipher = None
 
-
+himps = [
```
RemyDeWolf, line 7: can we sort this list for maintenance?
also it can be broken into 3 lists for each cloud providers

NitinAgg: Can we instead have a function which generates this list

bigbitbus: @RemyDeWolf @NitinAgg great suggestions! Addressed both  in new commit.

bigbitbus:
Passing tests here - https://app.circleci.com/pipelines/github/run-x/opta/514/workflows/293fb6de-d40d-46e8-9382-b485669a5d0b

Ignore the redis test - its the expiring aws eks credentials issue https://github.com/hashicorp/terraform/issues/29182

opta.spec
```diff
@@ -1,9 +1,53 @@
...
+        modname = filename.split(".py")[0]
+        himps.append(".".join(["modules", modname, modname]))
+  return sorted(himps)
+
+
 
 block_cipher = None
 
+himps = generate_module_import_array()
+# himps = [
```
NitinAgg, line 19: nit: remove these comments?

NitinAgg (approved):

opta.spec
bigbitbus: Done

PR #557, opened 2021-12-23
Now we will only verify semver if we have the old semver

juandiegopalomino:
# Description

This means we only show version upgrade warnings if we know what the old version is, as in the _verify_semver function the first lines are
```python
def _verify_semver(old_semver_string: str, current_semver_string: str) -> None:
    if old_semver_string in [DEV_VERSION, ""] or current_semver_string in [
        DEV_VERSION,
        "",
    ]:
        return

````

# Safety checklist

n/a

### DIFF
--- a/opta/commands/apply.py
+++ b/opta/commands/apply.py
@@ -199,10 +199,10 @@ def _apply(
     old_semver_string = (
         ""
         if existing_config is None
-        else existing_config.get("opta_version", "0.0.1").strip("v")
+        else existing_config.get("opta_version", "").strip("v")
     )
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: opta/commands/apply.py [hunk 1]
```diff
@@ -197,7 +197,7 @@ def _apply(
```
NitinAgg, line 197: Let's just remove this warning for now

juandiegopalomino: ... why? There is a good reason we added it, and removing it is jumpy

NitinAgg (approved):

PR #558, opened 2021-12-23
Hotfix/add old niceproc for nontf commands

bigbitbus:
# Description
Explicitly import modules when building opta binary since dynamic imports are not detected by pyinstaller.

# Safety checklist

Manually

### DIFF
--- a/opta/nice_subprocess.py
+++ b/opta/nice_subprocess.py
@@ -74,11 +74,13 @@ def nice_run(  # type: ignore # nosec
     exit_timeout: Optional[float] = None,
     tee: bool = True,
     check: bool = False,
-    use_new_nice_run=False,
+    use_asyncio_nice_run=False,
     **kwargs,
 ) -> CompletedProcess:
 
-    if use_new_nice_run is False:
+    if (
+        use_asyncio_nice_run is False
+    ):  # Just use the original blocking Popen, this doesn't log subprocess output to datadog
         if input is not None:
             if kwargs.get("stdin") is not None:
                 raise ValueError("stdin and input arguments may not both be used.")
--- a/opta/nice_subprocess.py
+++ b/opta/nice_subprocess.py
@@ -4,11 +4,25 @@ import signal
 import sys
 import tempfile
 from asyncio import TimeoutError
-from subprocess import CalledProcessError, CompletedProcess  # nosec
+from subprocess import (  # nosec
+    PIPE,
+    CalledProcessError,
+    CompletedProcess,
+    Popen,
+    TimeoutExpired,
+)
 from traceback import format_exc
 from typing import Optional, Union
 
 import psutil
+
+try:
+    import _winapi  # noqa
+
+    _mswindows = True
+except ModuleNotFoundError:
+    _mswindows = False
+
 
 from opta.constants import DEV_VERSION, VERSION
 from opta.datadog_logging import DatadogLogHandler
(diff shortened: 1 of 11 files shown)


juandiegopalomino (approved):
awesome

RemyDeWolf:
tested locally, issue fixed ✔️ 
```
((opta) ) ➜  getting-started git:(hotfix/add-old-niceproc-for-nontf-commands) ✗ python3 ../../opta/cli.py shell -c opta.yaml           
Updated context arn:aws:eks:us-east-1:248233625043:cluster/opta-staging in /Users/remy/.kube/config
Updated context arn:aws:eks:us-east-1:248233625043:cluster/opta-staging in /Users/remy/.kube/config
root@staging-hello-k8s-service-769f68ffcf-9ffx8:/app# 
root@staging-hello-k8s-service-769f68ffcf-9ffx8:/app# 
root@staging-hello-k8s-service-769f68ffcf-9ffx8:/app# 
root@staging-hello-k8s-service-769f68ffcf-9ffx8:/app# ls
Dockerfile  README.md  __pycache__  app.py
root@staging-hello-k8s-service-769f68ffcf-9ffx8:/app# pwd
/app
```

opta/nice_subprocess.py
```diff
@@ -60,65 +74,119 @@ def nice_run(  # type: ignore # nosec
     exit_timeout: Optional[float] = None,
     tee: bool = True,
     check: bool = False,
+    use_new_nice_run=False,
```
RemyDeWolf, line 77: `old/new` is unclear to me which one I would need to use if I'm writing a new module.

can we come up with a name that carries some meaning about when to use one or the other?
and some comments on this function so someone using this function will understand what are the differences and make a decision about which one to use.

bigbitbus: Good point, fixe by renaming flag and adding a comment.

opta/nice_subprocess.py
```diff
@@ -60,65 +74,121 @@ def nice_run(  # type: ignore # nosec
...
-    try:
-        listargs = list(popenargs)
-        listargs[0].insert(0, "exec")
-        popenargs = tuple(listargs)
-        log_to_datadog(
-            "Calling subprocess with these arguments:\n" + " ".join(*popenargs), "INFO"
-        )
+    if (
+        use_asyncio_nice_run is False
+    ):  # Just use the original blocking Popen, this doesn't log subprocess output to datadog
```
RemyDeWolf, line 83: can this be put in the Docstring so it doesn't require to look inside the function when calling it?

opta/nice_subprocess.py
```diff
@@ -60,65 +74,121 @@ def nice_run(  # type: ignore # nosec
...
 
-    try:
-        listargs = list(popenargs)
-        listargs[0].insert(0, "exec")
-        popenargs = tuple(listargs)
-        log_to_datadog(
-            "Calling subprocess with these arguments:\n" + " ".join(*popenargs), "INFO"
-        )
+    if (
+        use_asyncio_nice_run is False
```
RemyDeWolf, line 82: what about datadog logging? there is also a different behavior?
to JD's point, sometimes it's not desirable to log to datadog, the shell is a good example.
we could document that as well, or expose a specific boolean.
if asyncio && datadog logging are not supported together right now, we can also raise an exception so it would fail when testing it.

opta/nice_subprocess.py
bigbitbus: Yea we dont really have a convention around documenting functions, lets discuss that in the team and then we can add such knowledge into a standardized function docstring for example.

opta/nice_subprocess.py
bigbitbus: In the old nice process there was no datadog logging.

There is also a switch in the nice_subprocess function to turn off the logging when needed, again which functions need to be quieter needs to be discussed and then the code put into place. This ticket is only the hotfix for the bug.

opta/nice_subprocess.py
RemyDeWolf: ok, let's do that 👍

opta/nice_subprocess.py
RemyDeWolf: go it, let's do hotfix and discuss later

RemyDeWolf (approved):

NitinAgg (approved):

NitinAgg (approved):

PR #559, opened 2022-01-03
Initial work on layer v2

rabbitfang:
# Description

# Safety checklist

YOUR_ANSWER


bigbitbus (approved):
Lets merge this into the module-api branch.

PR #561, opened 2022-01-03
First version of the meister profiler

juandiegopalomino:
# Description

Contains the first version of the meister profiler/decorator. This is just a small package with a decorator, meister_timer, being the only thing (besides the config vars) someone should be setting. What this decorator does is simple: when a decorated function is called (works for sync and async), it prints the time it took and keeps track of the call's metadata (start time, end time, calling line, etc...) and at the end uses it to report a list of the top longest calls as well as a color-coded bar graph representing the rough times
Screen Shot 2022-01-03 at 1 30 03 PM

# Safety checklist

Ran it locally a bunch


opta/cli.py
```diff
@@ -61,6 +63,7 @@ def cli() -> None:
         # after the command.
         # However, we should still clean them up before the next command, or
         # else it may interfere with it.
+        meister.MEISTER_REPORT_ENABLED = VERSION == DEV_VERSION
```
NitinAgg, line 66: Might be better to have a separate flag for this

NitinAgg (commented):
Worried about adding this decorator everywhere. Would it be possible to add a visualizer on top of cprofile (context: https://stackoverflow.com/questions/582336/how-can-you-profile-a-python-script)

bigbitbus (commented):
We could create another repo for meister under runx and market it as a tool for the community - it will get some traffic to runx (although it will be generic python perf searches rather than IaC only). Who knows, it might just take off!

juandiegopalomino:
@NitinAgg believe me, I truly wish you were right, and doing something like that was the first thing I tried. Unfortunately I abandoned that approach because
1. Usually the profiler needed to be run as a wrapping command or weird flag
2. It did not give you the important metadata (e.g. start time, end time, caller, etc...)
3. (and this is what really killed it), those profilers checked **everything**. Every call to imported function, and every line of code. It was an overkill and trying to parse the data actually slowed down opta to a snail crawl

I agree with you that having it as a decorator in every function is tedious and it was a defeat I'm not happy with, but I couldn't think of a way around it. I don't think it was a major setback as we could in the end get more succinct, "quick"  metadata which we wanted for the purpose of increasing opta's speed. But if you do find a way to simply have a "magic one line that sets it all up" pls pls pls let me knw

juandiegopalomino:
@bigbitbus Yes we could, but I want to let it "incubate" a bit in opta to see if it truly is useful.

opta/meister.py
```diff
@@ -0,0 +1,172 @@
...
+    caller_line: int
+
+
+MEISTER_RECORDS: List[TimedCall] = []
+MEISTER_ASYNC_LOCK = Lock()
+MEISTER_REPORT_SIZE = 20
+MEISTER_TIME_THRESHOLD = 0.1
+MEISTER_REPORT_ENABLED = True
+BLOCK_SYMBOL = "\u2588"
+MEISTER_START_TIME = time.time()
```
rabbitfang, line 36:
```suggestion
ASYNC_LOCK: Final = Lock()
REPORT_SIZE: Final = 20
TIME_THRESHOLD: Final = 0.1
REPORT_ENABLED: bool = True
BLOCK_SYMBOL: Final = "\u2588"
START_TIME: Final = time.time()
```

- We already know these are constants/variables for meister, since we are in that module. No need to duplicate the module name in the variable names.
- Added `typing.Final` to some variables to hint that they are constant.

opta/meister.py
```diff
@@ -0,0 +1,172 @@
...
+            positional_args_str = positional_args_str[:36] + "...)"
+        keyword_args_str = str(record["keyword_args"])
+        if len(keyword_args_str) > 40:
+            keyword_args_str = keyword_args_str[:36] + "...)"
+        print(
+            f"{fg(color_idx)}{record['function_name']}{attr(0)} called in file {record['caller_file']}, line {record['caller_line']} with args {positional_args_str}, "
+            f"kwargs {keyword_args_str}: {round_sig(record['end_time'] - record['start_time'])} seconds"
+        )
+        color_map[record["id"]] = color_idx
+    max_column = os.get_terminal_size()[0]
```
rabbitfang, line 123:
```suggestion
    try:
        max_column = os.get_terminal_size()[0]
    except OSError:
        max_column = 80  # Assume 80 columns if we aren't in a terminal
```

`os.get_terminal_size` raises if stdout isn't attached to a terminal.

opta/cli.py
```diff
@@ -7,6 +7,7 @@
 from click_didyoumean import DYMGroup
 from colored import attr, fg
 
+import opta.meister as meister
```
rabbitfang, line 10:
```suggestion
from opta import meister
```

opta/cli.py
rabbitfang: Or an env var

rabbitfang (changes requested):
It's a cool feature. I wonder if we could find a way to automatically annotate functions that opta defines.

Pairing up the function colors with the lanes below might be hard to do. A suggestion would be to print the function's fully qualified name (or part of it; includes module name) inside of the bar graphs in white.

bigbitbus:
> 

Since the decorator is being added to most class methods, we could look at an automated way of adding this to the callables in a class (e.g. https://stackoverflow.com/questions/6307761/how-to-decorate-all-functions-of-a-class-without-typing-it-over-and-over-for-eac); that way we won't make so many code changes and it will be relatively easy to comment/uncomment this for code clarity.

NitinAgg:
> 3\. (and this is what really killed it), those profilers checked **everything**. Every call to imported function, and every line of code. It was an overkill and trying to parse the data actually slowed down opta to a snail crawl

@bigbitbus I was also looking into this. but unfortunately we don't use classes that much and even if we do we would still have to add the decorator to each class.

opta/utils/__init__.py
```diff
@@ -4,6 +4,7 @@
 import sys
 from logging import Formatter, Logger, LogRecord
 from logging.handlers import QueueHandler, QueueListener
+from os.path import exists
```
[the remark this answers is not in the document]
juandiegopalomino: because of mocking. If we did not set it like this, the old mocks would have mocked the actual os.path.exist function instead of the version imported in each module

opta/meister.py
```diff
@@ -0,0 +1,174 @@
...
+    end_time: float
+    caller_file: str
+    caller_line: int
+
+
+_records: List[TimedCall] = []
+ASYNC_LOCK: Final = Lock()
+REPORT_SIZE: Final = 20
+TIME_THRESHOLD: Final = 0.1
+REPORT_ENABLED: bool = True
```
rabbitfang, line 35:
```suggestion
REPORT_ENABLED: bool = False
```

This will avoid measuring functions when running unit tests.

rabbitfang (approved):
Minor change. Otherwise, LGTM

opta/cli.py
```diff
@@ -61,6 +63,7 @@ def cli() -> None:
         # after the command.
         # However, we should still clean them up before the next command, or
         # else it may interfere with it.
+        meister.REPORT_ENABLED = VERSION in [DEV_VERSION, ""]
```
NitinAgg, line 66: Let's make this based on an env flag instead. I don't want outside contributors to run into this.

NitinAgg (commented):
Minor comment

PR #562, opened 2022-01-04
fix typo in the info message to use --detailed-plan

RemyDeWolf:
# Safety checklist

untested, just a text change


bigbitbus (approved):
lgtm

PR #571, opened 2022-01-06
Added a class for mixed literal and interpolated substring parsing

bigbitbus:
# Description
Parser for mixed literalstrings + interpolated references

# Safety checklist

## How has this change been tested, beside unit tests
Not tested yet in prod code, part of module API branch


bigbitbus:
As discussed on slack, @rabbitfang suggested changes that look great, lets merge it.

PR #575, opened 2022-01-10
Destroy persistent volumes when destroying a service

RemyDeWolf:
# Description
Some context:
- Kubernetes API: _PersistentVolumes associated with the Pods' PersistentVolume Claims are not deleted when the Pods, or StatefulSet are deleted._ https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#stable-storage
- Helm is consistent with k8s, it doesn't delete PVC when deleting a StatefulSet https://github.com/helm/helm/issues/3313
- Currently Opta uses Helm to uninstall a service, so the persistent volumes are always kept.

How do developer usually solve this issue?
- Manually delete these resources (as suggested by k8s documentation)
- (or) Configure a `post-delete` [hook](https://helm.sh/docs/topics/charts_hooks/) in helm. [example](https://github.com/helm/helm/issues/5137)
- (or) Use this kubernetes 1.23 alpha feature: PersistentVolumeClaim retention https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention

How can we support it with Opta?
1. Configure a `post-delete` hook in helm.
    - Complexity with getting helm to run kubectl with the necessary permission, see https://github.com/helm/helm/issues/5137
2. Configure a `local-exec` in terraform running kubectl when destroy is invoked
    - This would become a shell script to maintain/test
3. Configure a opta `post_delete` hook when the k8s service module processor
    - This is would be a first-class citizen support with opta, python change

This PR implements 3.

# Safety checklist
* [N] This change will NOT lead to data loss
    - **Previously we would keep the PVC and PV, and the underlying storage (EBS, PersistentDisk) when doing `opta destroy` on a service file. With this change, destroy would delete the PVC and PV and their underlying storage**
    - cc @NitinAgg  Proposal (not implemented with this PR): If this needs to be backward compatible, we could have a new field in `persistent_storage` to define the claim retention policy (to mirror https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention)
ex:
```yaml
    persistent_storage:
      - path: "/my-pv"
        size: 1 # 1 GB
        retention_policy: [Retain/Delete] # default could be Retain for backward compatibility
```

- added new unit tests
- tested with AWS, EBS is deleted
- tested with GCP, Persistent Disk is deleted

Example of Opta destroy ouput:
```
Destroy complete! Resources: 5 destroyed.

Deleting persistent volumes in namespace 'hellopv'
Deleting persistent volume claim 'opta-persistent-0-hellopv-hellopv-k8s-service-0'
```

### DIFF
--- a/opta/core/kubernetes.py
+++ b/opta/core/kubernetes.py
@@ -439,7 +439,7 @@ def delete_persistent_volume_claims(
 
     claims = list_persistent_volume_claims(namespace=namespace, opta_managed=opta_managed)
     if not claims:
-        logger.info(
+        logger.debug(
             f"No persistent volume claim (opta_managed: {opta_managed}) found in namespace '{namespace}', skipping persistent volume cleanup"
         )
         return
--- a/opta/core/kubernetes.py
+++ b/opta/core/kubernetes.py
@@ -13,12 +13,14 @@ from kubernetes.client import (
     AppsV1Api,
     CoreV1Api,
     V1ConfigMap,
+    V1DeleteOptions,
     V1Deployment,
     V1DeploymentList,
     V1Event,
     V1Namespace,
     V1ObjectMeta,
     V1ObjectReference,
+    V1PersistentVolumeClaim,
     V1Pod,
     V1Secret,
     V1SecretList,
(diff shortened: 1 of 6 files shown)


juandiegopalomino (approved):

opta/core/kubernetes.py
```diff
@@ -360,6 +362,99 @@ def list_namespaces() -> None:
...
+    asynchronous HTTP request, please pass async_req=True
+
+    :param str namespace: namespace to search for the Persistent Volume Claims
+    :param bool opta_managed: filter to only delete objects managed by opta
+    :param bool async_req: execute request asynchronously
+    """
+
+    claims = list_persistent_volume_claims(namespace=namespace, opta_managed=opta_managed)
+    if not claims:
+        logger.info(
```
NitinAgg, line 442: this should be a `debug` log

NitinAgg (commented):
minor nit. But looks good!

NitinAgg (approved):

PR #577, opened 2022-01-11
Adding memo for obsolete code

juandiegopalomino:
# Description

# Safety checklist

n/a

### DIFF
--- a/opta/datadog_logging.py
+++ b/opta/datadog_logging.py
@@ -17,6 +17,12 @@ logger = getLogger("opta")
 
 
 class DatadogLogHandler(Handler):
+    """
+    NOTE: Kinda obsolete because an official version came out: https://datadog-api-client.readthedocs.io/en/latest/v2/HTTPLogItem/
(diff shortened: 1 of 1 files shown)


NitinAgg (approved):

opta/datadog_logging.py
```diff
@@ -17,6 +17,12 @@
 
 
 class DatadogLogHandler(Handler):
+    """
+    NOTE: Kinda obsolete because an official version came out: https://datadog-api-client.readthedocs.io/en/latest/v2/HTTPLogItem/
+    But we're getting all that we want from it for now and has proven really robust/some of our oldest (yet constantly used) code.
+    Go ahead and switch to the official package if there's like a new feature which we need. Otherwise why fix what's not broken?
+    """
+
```
RemyDeWolf, line 25: thanks for documenting this.
I also found this library that could be useful for future use https://pypi.org/project/Deprecated/

RemyDeWolf (approved):

PR #578, opened 2022-01-11
Updated the behaviour of opta apply and destroy if auto-approve flag is provided.

nsarupr:
# Description
If the `--auto-approve` flag is provided, then
`opta destroy` doesn't confirm the layer name with a User prompt, rather just gives a 2 second gap to the User to Abort.
`opta apply` in case of an existing deployment will keep the deployment alive.

# Safety checklist

Tested Manually.

### DIFF
--- a/opta/commands/destroy.py
+++ b/opta/commands/destroy.py
@@ -1,6 +1,5 @@
 import os
 import time
-from colored import attr, fg
 from pathlib import Path
 from typing import Dict, List, Optional
 
@@ -8,6 +7,7 @@ import boto3
 import click
 from azure.storage.blob import ContainerClient
 from botocore.config import Config
+from colored import attr, fg
 from google.cloud import storage  # type: ignore
 from google.cloud.exceptions import NotFound
 
--- a/opta/commands/destroy.py
+++ b/opta/commands/destroy.py
@@ -21,7 +21,7 @@ from opta.error_constants import USER_ERROR_TF_LOCK
 from opta.exceptions import UserErrors
 from opta.layer import Layer
 from opta.pre_check import pre_check
-from opta.utils import check_opta_file_exists, fmt_msg, logger
+from opta.utils import check_opta_file_exists, logger
 
 
 @click.command()
--- a/opta/commands/destroy.py
+++ b/opta/commands/destroy.py
@@ -90,13 +90,14 @@ def destroy(
 
     tf_flags: List[str] = []
     if auto_approve:
+        sleep_time = 2
         logger.info(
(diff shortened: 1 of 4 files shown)


juandiegopalomino (approved):

opta/commands/destroy.py
```diff
@@ -88,16 +90,13 @@ def destroy(
...
-                Are you REALLY sure you want to run destroy with auto-approve?
-                ~Please make sure *{layer.name}* is the correct opta config.
-                """
-            ),
-            abort=True,
+        logger.info(
+            f"{fg('red')}{attr('bold')}Opta will now destroy the {attr('underlined')}{layer.name}{attr(0)}"
+            f"{fg('red')}{attr('bold')} layer.{attr(0)}\n"
+            f"{fg('red')}{attr('bold')}Sleeping for {attr('underlined')}2 secs{attr(0)}"
+            f"{fg('red')}{attr('bold')}, if user wants to abort.{attr(0)}"
```
RemyDeWolf, line 97: why red? it is not an error, just an info message.
can you also tell the user how to exit ?
`Press Ctrl-C to abort now.`

opta/commands/destroy.py
```diff
@@ -88,16 +90,13 @@ def destroy(
...
         )
+        time.sleep(2)
```
RemyDeWolf, line 99: can 2 be a variable so it can be easily changed here and in the message?

RemyDeWolf (commented):
What do you think if we make this more generic?

```python
click_confirm(confirm_message: str, auto_approve: bool = false, pause_message: str, pause_sec: int = 0) 
    # if pause set: display pause_message for x sec, and wait
    # if auto_approve true, return
    # if auto_approve false, prompt user
```

review existing calls of `click.confirm()` to see which one to use.
we have only 7 calls, and I would think that they should honor `auto_approve`

opta/core/generator.py
```diff
@@ -59,11 +61,21 @@ def gen(
...
+                            f"did not specify an image tag. Would you like to keep the existing deployment alive?",
+                        ):
+                            image_tag = current_image_info["tag"]
+                            image_digest = current_image_info["digest"]
+                    else:
+                        logger.info(
+                            f"{fg('green')}{attr('bold')}Using the existing deployment {attr('underlined')}"
+                            f"(tag={current_image_info['tag']}, digest={current_image_info['digest']}).{attr(0)}\n"
+                            f"{fg('green')}{attr('bold')}If you wish to deploy another image, please use "
+                            f"{fg('green')}{attr('bold')}{attr('underlined')} opta deploy command.{attr(0)}"
```
RemyDeWolf, line 77: similar comment, no need for green error, default color is fine.

PR #579, opened 2022-01-11
Stateless TF Unlock for AWS and GCP

nsarupr:
# Description
A special case, where the TF Lock gets created even before the State is stored in the Buckets. 
For this we are forcefully removing the TF Lock.
- AWS: Manually deleting the entry created in `Dynamo DB`.
- GCP: Manually deleting the `.tflock` file created.

Thanks @juandiegopalomino  for pointing out the issue.

# Safety checklist

This has been manually tested.

### DIFF
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -153,6 +153,16 @@ class AWS(CloudClient):
         except Exception:
             return ""
 
+    def force_delete_terraform_lock_id(self) -> None:
+        bucket = self.layer.state_storage()
+        providers = self.layer.gen_providers(0)
+        dynamodb_table = providers["terraform"]["backend"]["s3"]["dynamodb_table"]
+
+        self.__get_dynamodb(dynamodb_table).delete_item(
+            TableName=dynamodb_table,
+            Key={"LockID": {"S": f"{bucket}/{self.layer.name}"}},
+        )
+
     @staticmethod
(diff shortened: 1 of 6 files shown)


juandiegopalomino (approved):
awesome

opta/core/aws.py
```diff
@@ -153,6 +153,16 @@ def get_terraform_lock_id(self) -> str:
         except Exception:
             return ""
 
+    def force_delete_terraform_lock_id(self) -> None:
```
RemyDeWolf, line 156: is it possible to add tests on this new functions?

opta/core/terraform.py
```diff
@@ -952,11 +952,26 @@ def force_unlock(cls, layer: "Layer", *tf_flags: str) -> None:
...
-        )
+        try:
+            nice_run(
+                ["terraform", "force-unlock", *tf_flags, lock_id],
+                use_asyncio_nice_run=True,
+                check=True,
+            )
+        except Exception as e:
+            logger.info(
+                "An exception occured while removing the Terraform Lock. Trying to Remove the lock forcefully."
```
RemyDeWolf, line 963: the part about `Trying to Remove the lock forcefully.` should only be logged if we are about to do it (aws/gcp), can it be moved inside that function?

RemyDeWolf (approved):

PR #580, opened 2022-01-11
logging fix

juandiegopalomino:
# Description
to be able to parse resource errors

# Safety checklist

manually verified


bigbitbus (approved):
lgtm

PR #585, opened 2022-01-13
Feature/kubernetes resource

bigbitbus:
# Description
This PR implements a module to apply a kubernetes manifest yaml file via opta.
Sample to run this
```
(opta) sachin@sachin-bigbitbus-iwl:~/alt/runx/opta$ cat testk8sresource.yaml 
name: testk8sresource
environments:
  - name: awsenv
    path: "awsenv.yaml"

modules:
  - name: pod
    type: k8s-manifest
    file_path: k8test.yaml
(opta) sachin@sachin-bigbitbus-iwl:~/alt/runx/opta$ cat k8test.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: nginx
  namespace: default
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
      protocol: TCP

```

# Safety checklist
* [ x] This change is backwards compatible and safe to apply by existing users
* [ x] This change will NOT lead to downtime who already has an env/service setup

Manual test


modules/k8s_resource/tf_module/main.tf
```diff
@@ -0,0 +1,13 @@
...
+  config_path    = var.kubeconfig
+  config_context = var.kubecontext
+}
+
+resource "kubernetes_manifest" "manifest" {
+  manifest = yamldecode(file(var.file_path))
+  timeouts {
+    update = "2m"
+    create = "5m"
+    delete = "2m"
```
RemyDeWolf, line 11: 2m, could be a bit low?
I wonder if we can use the same value for these 3 and support an optional module variable `timeout_min`.
just in case someone needs a higher timeout they could set it up from config file, and we don't need to change opta

modules/k8s_resource/tests/test_k8s_resource.py
```diff
@@ -0,0 +1,22 @@
...
+                os.getcwd(), "tests", "fixtures", "dummy_data", "local_dummy_config.yaml"
+            ),
+            None,
+        )
+        idx = len(layer.modules)
+        k8sresource_module = layer.get_module("k8sresource", idx)
+        K8sresourceProcessor(k8sresource_module, layer)
+        assert(k8sresource_module.data["kubeconfig"] == "~/.kube/config")
+        assert(k8sresource_module.data["kubecontext"] == "kind-opta-local-cluster")
+
```
RemyDeWolf, line 20: anyway that we can test that a TF resource `kubernetes_manifest` was created?

modules/k8s_resource/k8s-resource.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$id": "https://app.runx.dev/modules/k8s-resource",
+  "description": "Kubernetes Resource",
```
RemyDeWolf, line 3: I wonder if we should name this `k8s-manifest`
- a manifest file can define multiple resources, so `k8s-resource` (singular) could be misleading in that way
- consistent with terraform naming: better for terraform users
what do you think?

modules/k8s_resource/tf_module/variables.tf
```diff
@@ -0,0 +1,31 @@
+variable "file_path" {
+  type    = string
+  default = "k8test.yaml"
```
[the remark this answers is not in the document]
bigbitbus: Its not used, I'll remove it.

modules/k8s_resource/tf_module/main.tf
bigbitbus: 5m for each ok?
About the other variable, We could add it when a customer runs into 5 mins being too short, but its probably not needed right now (for example, when using kubectl we dont specify timeouts).

modules/k8s_resource/tests/test_k8s_resource.py
bigbitbus: We'll add a CI test for that in circle CI (e2e testing instead of the unit test here).

modules/k8s_resource/k8s-resource.json
bigbitbus: sure, will rename

modules/k8s_manifest/k8s-manifest.md
```diff
@@ -0,0 +1,8 @@
+---
+title: "kubernetes-resource"
+linkTitle: "kubernetes-resource"
```
RemyDeWolf, line 3: rename here too

modules/k8s_resource/tf_module/main.tf
RemyDeWolf: yes 5min for all 👍

modules/k8s_resource/k8s-resource.json
bigbitbus: done

modules/k8s_manifest/k8s-manifest.json
```diff
@@ -0,0 +1,35 @@
+{
+  "$id": "https://app.runx.dev/modules/k8s-manifest",
+  "description": "Kubernetes Resource",
```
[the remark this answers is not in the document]
bigbitbus: Done

NitinAgg (approved):

juandiegopalomino (changes requested):
Let's not use kubeconfig in the provider. Currently we do something like this:

```
    kubernetes:
      host: "${{{module_source}.k8s_endpoint}}"
      token: "${{data.aws_eks_cluster_auth.k8s.token}}"
      cluster_ca_certificate: "${{base64decode({module_source}.k8s_ca_data)}}"
```

bigbitbus:
> .k8s.token

I am looking at the documentation of the K8s TF provider here: https://registry.terraform.io/providers/hashicorp/kubernetes/latest/docs

If we want to avoid kubeconfig file then for this provider we need

```
  client_certificate     = file("~/.kube/client-cert.pem")
  client_key             = file("~/.kube/client-key.pem")
  cluster_ca_certificate = file("~/.kube/cluster-ca-cert.pem")
```
I can't see the client_certificate and client_key in the terraform state of my example aws cluster; are we storing those during opta apply?

bigbitbus:
> Let's not use kubeconfig in the provider. Currently we do something like this:
> 
> ```
>     kubernetes:
>       host: "${{{module_source}.k8s_endpoint}}"
>       token: "${{data.aws_eks_cluster_auth.k8s.token}}"
>       cluster_ca_certificate: "${{base64decode({module_source}.k8s_ca_data)}}"
> ```

@juandiegopalomino as discussed I've separated the config file path into a function. Can you greenlightt he review now?

juandiegopalomino (approved):

PR #587, opened 2022-01-14
Feature/secret bulk-update

RemyDeWolf:
# Description
Implement the **opta secret bulk-update** command.

```bash
cat secrets.env
FROM_FILE_SECRET1=val1
FROM_FILE_SECRET2=val2

opta secret bulk-update -c service.yaml secrets.env
Success

opta secret list -c hello.yaml  
FROM_FILE_SECRET1=val1
FROM_FILE_SECRET2=val2

```

Changes to existing commands:
- `opta apply` does not check for all secrets to exist before applying
- `secret list` prints in the  `.env` format

Confirming existing behaviors:
- `secret list` prints all values, including linked secrets
- `secret update` allow you to update secrets that are not listed in the yaml file

Code refactored:
- `core/kubernetes.py` 
  -  No longer aware of  _linked_ and _manual_  secrets. This removes some duplication.
- new: `core/secrets.py`
  - Expose helper function for _linked_ and _manual_ secrets.

# Safety checklist
  * Note: the `secret list` command is now printing the values but I'm not aware of anybody using this function for automated provisioning, 

Tested manually on local cluster + added new unit tests.


juandiegopalomino (approved):

NitinAgg (approved):
💯

PR #590, opened 2022-01-18
Removed unsupported clouds from Atlas Mongodb

bigbitbus:
# Description

We had included GCP and azure in the registry values for mongodb-atlas; but in reality they are not supported in the code.

# Safety checklist
* [  X] This change is backwards compatible and safe to apply by existing users

Mnaually tested AWS + Mongo


NitinAgg (approved):

bigbitbus:
Confirmed working in apply for mongodb-atlas provider.

PR #591, opened 2022-01-18
Fixing some annoying Azure things

juandiegopalomino:
# Description
1. AKS admin group object ids are no longer required
2. Get rid of annoying messages about credential verification
3. Cache credentials for significant speedup

# Safety checklist

Manually tested

### DIFF
--- a/tests/core/test_azure.py
+++ b/tests/core/test_azure.py
@@ -1,3 +1,4 @@
+from typing import Generator
 from unittest.mock import Mock
 
 from azure.identity import DefaultAzureCredential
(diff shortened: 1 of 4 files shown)


NitinAgg (approved):

tests/core/test_azure.py
```diff
@@ -54,9 +54,11 @@ class TestAzure:
     def test_get_credentials(self, mocker: MockFixture) -> None:
         mocked_default_creds = mocker.patch("opta.core.azure.DefaultAzureCredential")
         Azure.get_credentials()
+        Azure.credentials = None
         mocked_default_creds.assert_called_once_with()
 
     def test_get_remote_config(self, mocker: MockFixture, azure_layer: Mock) -> None:
+        Azure.credentials = None
```
RemyDeWolf, line 61: can we use [this](https://stackoverflow.com/a/22638709) instead of having this code in every test?

PR #594, opened 2022-01-19
Adding documentation about how db creds are passed.

juandiegopalomino:
# Description
Just adding some docs explaining the security behind the links

# Safety checklist

n/a


modules/aws_mysql/aws-mysql.md
```diff
@@ -41,3 +41,10 @@ links:
 ```
 
 If present, this map must have renames for all 4 fields.
+
+These values are passed securely into your environment by using a kubernetes secret created by opta within you
```
NitinAgg, line 45: you -> your

NitinAgg (commented):
We need to also give the command to view this secret if they want to use it elsewhere

NitinAgg (approved):

PR #596, opened 2022-01-19
Feature opta upgrade

nsarupr:
# Description
Introducing `opta upgrade`.
This will help you upgrade your existing opta hassle free.

# Safety checklist

Manually tested.

### DIFF
--- a/opta/commands/upgrade.py
+++ b/opta/commands/upgrade.py
@@ -1,0 +1,52 @@
+import os
+
+import click
+import requests
+from colored import attr, fg
+
+from opta.constants import OPTA_INSTALL_URL, successfull_upgrade
+from opta.nice_subprocess import nice_run
+from opta.upgrade import check_version_upgrade
+from opta.utils import logger
+
+TEMP_INSTALLATION_FILENAME = "opta_installation.sh"
+
+
+def _make_installation_file() -> None:
+    logger.debug(f"Querying {OPTA_INSTALL_URL}")
+    resp = requests.get(OPTA_INSTALL_URL)
+    resp.raise_for_status()
+    with open(TEMP_INSTALLATION_FILENAME, "w") as file:
+        file.write(resp.text)
+    nice_run(["chmod", "777", TEMP_INSTALLATION_FILENAME])
+
+
+def _upgrade_successfull() -> None:
+    open(successfull_upgrade, "w").close()
+
+
+def _cleanup_installation_file() -> None:
+    if os.path.isfile(TEMP_INSTALLATION_FILENAME):
+        os.remove(TEMP_INSTALLATION_FILENAME)
+
+
+@click.command()
+def upgrade() -> None:
+    """
+    Upgrade Opta to the Latest version available
+    """
+    try:
+        upgrade_present = check_version_upgrade(is_upgrade_call=True)
+        if upgrade_present:
+            _make_installation_file()
+            os.system(f"yes 2>/dev/null | ./{TEMP_INSTALLATION_FILENAME}")
+            _upgrade_successfull()
+    except Exception:
+        logger.error(
+            f"{fg('red')}"
+            "\nUnable to Install Latest Version of Opta."
+            "\nPlease follow the Instructions on https://docs.opta.dev/installation"
+            f"{attr(0)}"
+        )
+    finally:
+        _cleanup_installation_file()
--- a/opta/commands/upgrade.py
+++ b/opta/commands/upgrade.py
@@ -4,6 +4,7 @@ import click
 import requests
 from colored import attr, fg
 
+from opta.nice_subprocess import nice_run
 from opta.constants import OPTA_INSTALL_URL, successfull_upgrade
 from opta.nice_subprocess import nice_run
 from opta.upgrade import check_version_upgrade
@@ -39,7 +40,7 @@ def upgrade() -> None:
         upgrade_present = check_version_upgrade(is_upgrade_call=True)
         if upgrade_present:
             _make_installation_file()
-            os.system(f"yes 2>/dev/null | ./{TEMP_INSTALLATION_FILENAME}")
+            nice_run([f"./{TEMP_INSTALLATION_FILENAME}"], input=b"y")
             _upgrade_successfull()
     except Exception:
         logger.error(
--- a/opta/commands/upgrade.py
+++ b/opta/commands/upgrade.py
@@ -4,7 +4,6 @@ import click
 import requests
 from colored import attr, fg
 
-from opta.nice_subprocess import nice_run
 from opta.constants import OPTA_INSTALL_URL, successfull_upgrade
 from opta.nice_subprocess import nice_run
 from opta.upgrade import check_version_upgrade
@@ -22,7 +21,7 @@ def _make_installation_file() -> None:
     nice_run(["chmod", "777", TEMP_INSTALLATION_FILENAME])
 
 
-def _upgrade_successfull() -> None:
+def _upgrade_successful() -> None:
     open(successfull_upgrade, "w").close()
 
 
@@ -41,7 +40,7 @@ def upgrade() -> None:
         if upgrade_present:
             _make_installation_file()
             nice_run([f"./{TEMP_INSTALLATION_FILENAME}"], input=b"y")
-            _upgrade_successfull()
+            _upgrade_successful()
     except Exception:
         logger.error(
             f"{fg('red')}"
--- a/opta/commands/upgrade.py
+++ b/opta/commands/upgrade.py
@@ -33,7 +33,7 @@ def _cleanup_installation_file() -> None:
 @click.command()
 def upgrade() -> None:
     """
-    Upgrade Opta to the Latest version available
+    Upgrade Opta to the latest version available
     """
     try:
         upgrade_present = check_version_upgrade(is_upgrade_call=True)
@@ -44,8 +44,8 @@ def upgrade() -> None:
     except Exception:
         logger.error(
             f"{fg('red')}"
-            "\nUnable to Install Latest Version of Opta."
-            "\nPlease follow the Instructions on https://docs.opta.dev/installation"
+            "\nUnable to install latest version of Opta."
+            "\nPlease follow the instructions on https://docs.opta.dev/installation"
             f"{attr(0)}"
         )
     finally:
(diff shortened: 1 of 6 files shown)


opta/commands/upgrade.py
```diff
@@ -0,0 +1,52 @@
...
+
+def _cleanup_installation_file() -> None:
+    if os.path.isfile(TEMP_INSTALLATION_FILENAME):
+        os.remove(TEMP_INSTALLATION_FILENAME)
+
+
+@click.command()
+def upgrade() -> None:
+    """
+    Upgrade Opta to the Latest version available
```
RemyDeWolf, line 36: tiny one, Latest -> latest

opta/commands/upgrade.py
```diff
@@ -0,0 +1,52 @@
...
+    try:
+        upgrade_present = check_version_upgrade(is_upgrade_call=True)
+        if upgrade_present:
+            _make_installation_file()
+            nice_run([f"./{TEMP_INSTALLATION_FILENAME}"], input=b"y")
+            _upgrade_successful()
+    except Exception:
+        logger.error(
+            f"{fg('red')}"
+            "\nUnable to Install Latest Version of Opta."
```
RemyDeWolf, line 47: capitalize is fine for title but for error message, let's use standard capitalization: -> _Unable to install latest version of Opta_

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
-                f"You have {VERSION} installed. Latest version is {latest_version}.\n"
-                f"Upgrade instructions are available at {UPGRADE_INSTRUCTIONS_URL}."
-            )
-    except Exception as e:
-        logger.debug(f"Semver check failed with error {e}")
+    if Path(successfull_upgrade).is_file():
+        os.remove(successfull_upgrade)
+        return False
+    if is_upgrade_call or _should_check_for_version_upgrade():
+        logger.info("Checking for version upgrades...")
```
RemyDeWolf, line 41: surprisingly when I test on main branch, this message was displayed twice sometimes:
```
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade                               
Checking for version upgrades...
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade --help   
Usage: cli.py upgrade [OPTIONS]

  Upgrade Opta to the Latest version available

Options:
  -h, --help  Show this message and exit.
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade       
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...
(.venv) ➜  opta git:(feat/nilesh/runx-1037-opta-upgrade) ✗ python3 $PYTHONPATH/opta/cli.py upgrade 
Checking for version upgrades...

```

with debug flag
```
python3 $PYTHONPATH/opta/cli.py upgrade 
INFO: Checking for version upgrades...
DEBUG: Querying https://dev-runx-opta-binaries.s3.amazonaws.com/latest
DEBUG: Semver check failed with error  is not valid SemVer string
INFO: Checking for version upgrades...
DEBUG: Querying https://dev-runx-opta-binaries.s3.amazonaws.com/latest
DEBUG: Semver check failed with error  is not valid SemVer string
```

probably this call https://github.com/run-x/opta/blob/main/opta/cli.py#L86

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
+        try:
+            latest_version = _get_latest_version()
+        except Exception as e:
+            logger.debug(e, exc_info=True)
+            logger.debug("Unable to find latest version.")
```
RemyDeWolf, line 46: this could be info otherwise the user has no feedback on what was wrong

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
+                )
+                if not is_upgrade_call:
+                    print(
+                        f"Upgrade instructions are available at {UPGRADE_INSTRUCTIONS_URL}  or simply use the `opta upgrade` command"
+                    )
+                return True
+            else:
+                logger.info("User on the Latest Version.")
+        except Exception as e:
+            logger.debug(f"Semver check failed with error {e}")
```
RemyDeWolf, line 62: this could be info otherwise the user has no feedback on what was wrong

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
-        if semver.VersionInfo.parse(VERSION.strip("v")).compare(latest_version) < 0:
-            logger.warning(
-                "New version available.\n"
-                f"You have {VERSION} installed. Latest version is {latest_version}.\n"
-                f"Upgrade instructions are available at {UPGRADE_INSTRUCTIONS_URL}."
-            )
-    except Exception as e:
-        logger.debug(f"Semver check failed with error {e}")
+    if Path(successfull_upgrade).is_file():
+        os.remove(successfull_upgrade)
```
RemyDeWolf, line 38: I am not sure I follow this one, why is using a local file needed?

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
+    if is_upgrade_call or _should_check_for_version_upgrade():
+        logger.info("Checking for version upgrades...")
+        try:
+            latest_version = _get_latest_version()
+        except Exception as e:
+            logger.debug(e, exc_info=True)
+            logger.debug("Unable to find latest version.")
+            return False
+        try:
+            if semver.VersionInfo.parse(VERSION.strip("v")).compare(latest_version) < 0:
```
RemyDeWolf, line 49: this code is very similar to https://github.com/run-x/opta/blob/main/opta/commands/apply.py#L344-L350
can you have a common function to compare sem versions could opportunity to have unit test for it

opta/upgrade.py
```diff
@@ -25,28 +27,37 @@ def _get_latest_version() -> str:
...
+                    "New version available.\n"
+                    f"You have {VERSION} installed. Latest version is {latest_version}."
+                )
+                if not is_upgrade_call:
+                    print(
+                        f"Upgrade instructions are available at {UPGRADE_INSTRUCTIONS_URL}  or simply use the `opta upgrade` command"
+                    )
+                return True
+            else:
+                logger.info("User on the Latest Version.")
```
RemyDeWolf, line 60: Latest Version -> latest version

opta/commands/upgrade.py
```diff
@@ -0,0 +1,52 @@
...
+    open(successfull_upgrade, "w").close()
+
+
+def _cleanup_installation_file() -> None:
+    if os.path.isfile(TEMP_INSTALLATION_FILENAME):
+        os.remove(TEMP_INSTALLATION_FILENAME)
+
+
+@click.command()
+def upgrade() -> None:
```
RemyDeWolf, line 34: do you think we can add some unit test to test upgrade()?
this new part is not covered by existing tests right?

nsarupr: added

opta/upgrade.py
nsarupr: I added this one so as to not execute the check_version functionality if the update happens successfully.
Created a local file rather that setting up the Environment variable, because the implementation of the `os.unsetenv` differs for Linux and OSX machines

opta/upgrade.py
nsarupr: Semvers are manually set by us when releasing new versions, so we would not want to show the user this Error.

opta/upgrade.py
nsarupr: This happens only when the Version.txt file doesn't contain the correct Version. In that case, the upgrade itself won't happen. But we don't need to worry about that as of now, because we are maintaining the versions manually.

If you wish to test `opta upgrade`, please add correct version to the version.txt file. You will find the expected results.
And the reason you are sometimes getting the `Checking for version upgrades` because version check only happens in 20% of the runs.

RemyDeWolf: thank you for the explanation, I didn't know about this, where is the 20% rule defined in the code?

opta/upgrade.py
RemyDeWolf: is the issue to make sure that `check_version_upgrade` is only called once?
using a local file to save some state can be seen as [a side effect](https://en.wikipedia.org/wiki/Side_effect_(computer_science)).
would something like [this](https://stackoverflow.com/a/29764451) work?

opta/upgrade.py
RemyDeWolf: `Exception` is a generic error though, if this code evolve it might throw another exception that could be relevant to the user.
unless, this will create an error that would become verbose, it's ok to show it instead of failing silently.

opta/upgrade.py
nsarupr: Used [this](https://stackoverflow.com/a/29764379) to remove the Side Effect

opta/upgrade.py
nsarupr: Here you go: [LOC](https://github.com/run-x/opta/blob/main/opta/upgrade.py#L15)

RemyDeWolf (approved):
looks good!

PR #602, opened 2022-01-19
Improve Opta CLI documentation

RemyDeWolf:
# Description
- small CLI documentation improvements. Short title, brief description, provide consistent examples.
- renamed `opta init` to `opta ui`
- added links to documentation when useful

before:
```
Commands:
  apply              Initialize your environment or service to match the...
  configure-kubectl  Configure the kubectl CLI tool for the given cluster
  deploy             Push your new image to the cloud and deploy it in...
  destroy            Destroy all opta resources from the current config
  force-unlock       Force Unlocks a stuck lock on the current workspace
  init               Opens the interactive UI
  logs               Get stream of logs from your service
  secret             Commands for manipulating secrets for a k8s service
  shell              Get a bash shell into one of the pods in your service
```
after
```
  apply              Create or update infrastructure
  configure-kubectl  Configure kubectl so you can connect to the cluster
  deploy             Deploy a local image to Kubernetes
  destroy            Destroy all opta resources from the current config
  force-unlock       Release a stuck lock on the current workspace
  logs               Get stream of logs for a service
  secret             Manage secrets for a service
  shell              Get a shell into one of the pods in a service
  ui                 Opens the interactive UI
```

# Safety checklist
* [X] This change is backwards compatible and safe to apply by existing users: 
  * Note: **the command `opta init` is now `opta ui`**. But that should not be an issue as this command is not used for automation.

Ran `opta --help` for each command locally


NitinAgg (approved):
Love this! thanks

PR #604, opened 2022-01-20
Downgrade helm-chart version for admission controller to work in EKS

bigbitbus:
# Description
The Version 2.28.13 does not support admission controller on EKS. So we choose a version 2.10.3 from March 2021 that seems to work. Could it be a version between the two where things break? Definitely; since there are dozens of releases we'll go with 2.10.3 for now and determine where the issue lies before upgrading.

# Safety checklist

Manual testing

### DIFF
--- a/modules/datadog/tf_module/main.tf
+++ b/modules/datadog/tf_module/main.tf
@@ -17,7 +17,7 @@ resource "helm_release" "datadog" {
   repository = "https://helm.datadoghq.com"
   chart      = "datadog"
   name       = "${var.layer_name}-${var.module_name}"
-  version    = "2.28.13"
+  version    = "2.10.3"
   values = [
     yamlencode({
       datadog : {


NitinAgg (approved):

legalosLOTR:
Are we planning to merge this PR? Ideally, we should debug what's wrong with the latest version instead of downgrading.

bigbitbus:
> Are we planning to merge this PR? Ideally, we should debug what's wrong with the latest version instead of downgrading.

Yes, I have created a ticket in jira for RC. https://run-x.atlassian.net/browse/RUNX-1091; we know its the admission controller component of the cluster agent; dont know why its failing in the recent helm chart. So the new ticket covers

1. Determine where (which exact version) the helm chart becomes unusable
2. Compare the diffs of the helm chart version to try to Figure out the issue RC
3. Create GH issue and followup

legalosLOTR:
sounds good!

bigbitbus:
Confirmed worked for @legalosLOTR repro cluster, merging for now with the larger RC finding tbd in  https://run-x.atlassian.net/browse/RUNX-1091

PR #606, opened 2022-01-20
Refactoring special_formatter

nsarupr:
# Description
Refactoring the special_formatter to show logs when interpolations keys are incorrect.

# Safety checklist

Tested manually.
Sample Configuration:
```
environments:
  - name: aws-example
    path: "./aws.yml"
    variables:
      max_containers: 2
name: http-service
modules:
  - name: app
    type: k8s-service
    image: kennethreitz/httpbin
    public_uri: "http.{parent.omain}".              # Invalid key parent.omain
    min_containers: 2
    max_containers: "{vars.max_container}"          # Invalid key vars.max_container
    liveness_probe_path: "/get"
    readiness_probe_path: "/get"
    sticky_session: true
    sticky_session_max_age: 20
    keep_path_prefix: true
    port:
      http: 80
```

### DIFF
--- a/opta/special_formatter.py
+++ b/opta/special_formatter.py
@@ -1,12 +1,15 @@
 import logging
 import string
-from typing import Any
+from typing import Any, List
+
+from opta.exceptions import UserErrors
 
 logger = logging.getLogger("opta")
 
 
 class PartialFormatter(string.Formatter):
     def __init__(self, missing: str = "~~", bad_fmt: str = "!!") -> None:
+        self.bad_field_names: List[str] = list()
         self.missing, self.bad_fmt = missing, bad_fmt
 
     def get_field(self, field_name: str, args: Any, kwargs: Any) -> Any:
@@ -14,6 +17,7 @@ class PartialFormatter(string.Formatter):
         try:
             val = super().get_field(field_name, args, kwargs)
         except (KeyError, AttributeError):
+            self.bad_field_names.append(field_name)
             logger.debug(
                 f"Did not find field {field_name} when trying to format. Odds are "
                 "this is just setting it to an empty default and you can ignore this"
@@ -32,3 +36,10 @@ class PartialFormatter(string.Formatter):
                 raise
             else:
                 return self.bad_fmt
+
+    def is_valid(self) -> None:
+        log: str = "The following field names are not valid:"
+        if len(self.bad_field_names) > 0:
+            for bad_field_name in self.bad_field_names:
+                log += f"\n\t * {bad_field_name}"
+            raise UserErrors(log)
--- a/opta/special_formatter.py
+++ b/opta/special_formatter.py
@@ -1,15 +1,14 @@
 import logging
 import string
-from typing import Any, List
+from typing import Any
 
-from opta.exceptions import UserErrors
+from opta.constants import InterpolationErrors
 
 logger = logging.getLogger("opta")
 
 
 class PartialFormatter(string.Formatter):
     def __init__(self, missing: str = "~~", bad_fmt: str = "!!") -> None:
-        self.bad_field_names: List[str] = list()
         self.missing, self.bad_fmt = missing, bad_fmt
 
     def get_field(self, field_name: str, args: Any, kwargs: Any) -> Any:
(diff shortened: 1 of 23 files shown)


RemyDeWolf (commented):
my main concern with this approach is that it created a new class (Interpolation) that is tightly coupled to the PartialFormatter but in a different package.

I was looking at some open source code on how they solve this and I found this: 

1) the get field method raise an exception if a field is not found
https://github.com/django/django/blob/dc8bb35e39388d41b1f38b6c5d0181224e075f16/django/db/models/options.py?q=get_field#L583-L606

2) the part doing the validation would iterate over all fields and compound the exceptions in one when validation is over 
https://github.com/django/django/blob/dc8bb35e39388d41b1f38b6c5d0181224e075f16/django/db/models/base.py?q=FieldDoesNotExist#L516-L536

Can you see a similar approach for us would work?
I think it's actually not too far from what you have except that they use exception instead of adding to a set. It's up to the caller to decide how they did with the exception, they don't need to call a method (validate) later to know if something was wrong.
 if there is a need for a validate method, it could live next to the PartialFormatter.

opta/utils/globals.py
```diff
@@ -0,0 +1,38 @@
...
+
+from opta.exceptions import UserErrors
+
+
+class Interpolation:
+    bad_field_names: Set[str] = set()
+
+    @classmethod
+    def validate(cls) -> None:
+        if len(cls.bad_field_names) > 0:
```
rabbitfang, line 13:
```suggestion
        if not cls.bad_field_names:
            return
```

Then unindent the remaining method lines. Sets are falsey when empty (as are most python collection types), and by returning early, it reduces the amount of indentation for later lines in the method, known as a "guard clause"

opta/special_formatter.py
```diff
@@ -14,6 +16,7 @@ def get_field(self, field_name: str, args: Any, kwargs: Any) -> Any:
         try:
             val = super().get_field(field_name, args, kwargs)
         except (KeyError, AttributeError):
+            Interpolation.add(field_name)
```
rabbitfang, line 19: I really don't like the use of global state here. Use of globals makes reasoning about the world and logic harder. I would prefer if we bubble up from the hydrate method any issues, or pass down a "missing keys" object from the caller to keep everything within the call stack.

rabbitfang (changes requested):

opta/utils/globals.py
```diff
@@ -0,0 +1,38 @@
...
+            error = "Following field names could not be interpolated:"
+            for bad_field_name in cls.bad_field_names:
+                error += f"\n  * {bad_field_name}"
+
+            raise UserErrors(f"{fg('red')}{error}{attr(0)}")
```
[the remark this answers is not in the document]
nsarupr: First approach which I had was limiting the validations to the PartialFormatter only, but that didn't help the cause `hydrate` method is being called multiple times with different inputs, which leads to duplicate logs for a single incorrect key. That's why I kept this aggregation of all the field names in a global variable and then checking that global variable once the Terraform files have been generated for an iteration.

opta/special_formatter.py
nsarupr: Tried working at it from the hydrate as well, passing down the PartialFormatter with aggregated list of incorrect field names, but that way we would either end up throwing error in an iterative fashion for each incorrect key, and aggregating those errors using PartialFormatter wasn't much help because `hydrate` is called at multiple places with different inputs which would end up just giving partial list of incorrect keys.

opta/utils/globals.py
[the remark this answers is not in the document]
nsarupr: And yes, `validate` is sort of off putting, and I would change the naming convention for the same.

opta/special_formatter.py
RemyDeWolf: what's the next step here?

what if we just log a warning message here when the key is missing?
Worst case for the user the warning is frequent, but if they want to get rid of it they can fix their opta file. fair enough?

cc: @rabbitfang @nsarupr

nsarupr: @RemyDeWolf 
That can be done. And would be just a small change. We already add a debug log for the same.

rabbitfang (approved):

opta/special_formatter.py
RemyDeWolf: @nsarupr if this is a small change do you want to do it?
I know it's been approved, but if you think we can do better and it's not too much work, we can do that.

nsarupr: @RemyDeWolf @rabbitfang @juandiegopalomino @NitinAgg

RemyDeWolf (approved):

PR #608, opened 2022-01-21
Feature/local fixes

bigbitbus:
# Description

Several updates to opta local

1. Added support for opta's other commands in local environment:
events, shell, logs, force_unlock, secrets, configure_kubectl, output

2. Added the ability to optionally add "org_name" within (local) service file; so now, there is absolutely no need for environment opta yaml file; this means we can streamline the local documentation without confusion.
Example:
```
name: helloworld
org_name: hifi
modules:
  - name: firstapp
    type: k8s-service
    port:
      http: 80
    image: docker.io/kennethreitz/httpbin:latest
    healthcheck_path: "/get"
    links:
      - sessionstore
      - database
    public_uri: "/"
  - name: sessionstore
    type: redis    
  - name: database
    type: postgres
```
3. Moved the automatically generated local files from the PWD to ~/.opta/local/<org_name>/; so no pollution of user's directory.
4. Several other code and speed improvements 

# Safety checklist

Manual tests

### DIFF
--- a/opta/commands/deploy.py
+++ b/opta/commands/deploy.py
@@ -1,12 +1,9 @@
-import os
-from pathlib import Path
 from typing import Optional
 
 import click
-
+from opta.commands.apply import _local_setup
 from opta.amplitude import amplitude_client
 from opta.commands.apply import _apply
-from opta.commands.local_flag import _clean_tf_folder, _handle_local_flag
 from opta.commands.push import _push, is_service_config
 from opta.core.terraform import Terraform
 from opta.error_constants import USER_ERROR_TF_LOCK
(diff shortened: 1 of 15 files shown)


bigbitbus:
@RemyDeWolf @NitinAgg Going to be working on the doc changes with this code change.

opta/commands/deploy.py
```diff
@@ -81,6 +78,8 @@ def deploy(
     pre_check()
 
     config = check_opta_file_exists(config)
+    if local:
+        config = _local_setup(config, image_tag=tag, refresh_local_env=True)
```
RemyDeWolf, line 82: now that local is supported with all commands, would that make sense for `check_opta_file_exists(config)` to manage the local config and remove the logic from the caller?

```
    if local:
        config = _local_setup(config, image_tag=tag, refresh_local_env=True)
```
would become
```
    # local could be an optional parameter 
    config, env = check_opta_file_exists(config, local)
```

opta/commands/output.py
```diff
@@ -14,10 +15,19 @@
...
     "-e", "--env", default=None, help="The env to use when loading the config file"
 )
-def output(config: str, env: Optional[str],) -> None:
+@click.option(
+    "--local",
+    is_flag=True,
+    default=False,
+    help="""Use the local Kubernetes cluster for development and testing, irrespective of the environment specified inside the opta service yaml file""",
+    hidden=False,
+)
```
RemyDeWolf, line 24: it looks like it's possible to create a option once and reuse it in multiple commands: https://github.com/Carbyne/NOG5Thesis/blob/b78b03ac9ba337eabc3e3c3d479a4edd07260f49/netoglyc5/nog5/nog5/cli.py#L17

RemyDeWolf (approved):

PR #614, opened 2022-01-25
Preserve client ip fix

juandiegopalomino:
# Description
Making sure X-Forwarded-For is valid

# Safety checklist

manually

### DIFF
--- a/modules/aws_k8s_base/tf_module/ingress_nginx.tf
+++ b/modules/aws_k8s_base/tf_module/ingress_nginx.tf
@@ -18,7 +18,7 @@ resource "helm_release" "ingress-nginx" {
         config : local.config
         podAnnotations : {
           "linkerd.io/inject" : "enabled"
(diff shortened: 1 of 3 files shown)


modules/aws_k8s_base/tf_module/ingress_nginx.tf
```diff
@@ -18,6 +18,7 @@ resource "helm_release" "ingress-nginx" {
         config : local.config
         podAnnotations : {
           "linkerd.io/inject" : "enabled"
+          "config.linkerd.io/skip-inbound-ports": "80,443"
```
RemyDeWolf, line 21: 1. can we document that this should be removed with linkerd 2.12 include https://github.com/linkerd/linkerd2/issues/4219 for reference

2. can we document where appropriate to check https://linkerd.io/2.11/tasks/using-ingress/#nginx whenever we upgrade linkerd version, so we would review that the annotation we use are still valid

RemyDeWolf (approved):

PR #617, opened 2022-01-25
Fixing sticky sessions

juandiegopalomino:
# Description
https://linkerd.slack.com/archives/C89RTCWJF/p1643126236682619

Yestedar, we noticed that the origin IP was being lost due to a change a while back We figured out that it was a bug from linkerd and that the fix was to add the "config.linkerd.io/skip-inbound-ports" : "80,443" annotation to our nginx pods. Along the way, due to the explicit instructions on the linkerd documentation, I decided to add the nginx.ingress.kubernetes.io/service-upstream: "true" annotation. This was all done in this pr.

The problem is that now sticky sessions will not work and that extra annotation is needed for a linkerd feature which we did not need.

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/config/opta-k8s-service-helm/templates/ingress.yaml
+++ b/config/opta-k8s-service-helm/templates/ingress.yaml
@@ -28,7 +28,6 @@ metadata:
     {{- end }}
     cert-manager.io/cluster-issuer: opta-selfsigned
     kubernetes.io/ingress.class: "nginx"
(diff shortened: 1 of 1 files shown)


legalosLOTR (changes requested):
Let’s add some detailed context on this PR before merging

legalosLOTR (approved):

PR #621, opened 2022-01-25
Split destroy to plan and apply

nsarupr:
# Description

# Safety checklist

YOUR_ANSWER


opta/commands/destroy.py
```diff
@@ -105,8 +117,36 @@ def destroy(
     layer.modules = [x for x in layer.modules if x.name in modules]
     gen_all(layer)
     Terraform.init(False, "-reconfigure", layer=layer)
-    logger.info(f"Destroying {layer.name}")
-    Terraform.destroy_all(layer, *tf_flags)
```
juandiegopalomino, line 109: should we get rid of destroy_all?

nsarupr: I thought it better to have the `plan` and `apply` in `destroy.py` just the way it was in `apply.py`, and replicated the structure but with `-destroy` flag.
Also the state deletion which is happening inside `destroy_all` has been made as a separate method.
Is there anything that I might have missed?

juandiegopalomino: What I meant is that I don't think that the destroy_all function is used for anything else, so we should delete it

nsarupr: My bad. I understood different, but yeah, we should get rid of the `destroy_all` method.

juandiegopalomino (approved):

PR #622, opened 2022-01-25
Refactoring kubeconfig setup and adding kubeconfig purge

juandiegopalomino:
# Description
This change does 2 things:
1. The old configure kubectl method was poorly named, especially since now it did not actually change kubectl behavior. It has been renamed to set_kube_config and given a docstring explaining its purpose as creatign a separate kubeconfig file just for the current layer, which then may or may not be loaded to the default kubeconfig used by kubectl
2. Added a purge kube config function which cleans up the artificially created kubeconfig, usually as part of a full refresh. This can be done manually via the new --force flag of `opta configure-kubectl` and done automatically when a k8s cluster is deleted.

# Safety checklist

n/a

### DIFF
--- a/modules/aws_eks/aws_eks.py
+++ b/modules/aws_eks/aws_eks.py
@@ -9,6 +9,7 @@ from mypy_boto3_ec2.type_defs import NetworkInterfaceTypeDef
 from mypy_boto3_logs import CloudWatchLogsClient
 
 from modules.base import ModuleProcessor
+from opta.core.kubernetes import purge_opta_kube_config
 from opta.exceptions import UserErrors
 from opta.utils import logger
 
@@ -31,6 +32,7 @@ class AwsEksProcessor(ModuleProcessor):
         self.cleanup_cloudwatch_log_group(region)
         self.cleanup_dangling_enis(region)
         self.cleanup_security_groups(region)
+        purge_opta_kube_config(layer=self.layer)
 
     def cleanup_security_groups(self, region: str) -> None:
         logger.debug("Seeking dangling security groups EKS forgot to destroy.")
--- a/opta/commands/kubectl.py
+++ b/opta/commands/kubectl.py
@@ -16,14 +16,7 @@ from opta.utils import check_opta_file_exists
 @click.option(
     "-e", "--env", default=None, help="The env to use when loading the config file"
 )
-@click.option(
-    "-f",
-    "--force",
-    is_flag=True,
-    default=False,
-    help="Purge the existing cached kube config",
-)
-def configure_kubectl(config: str, env: Optional[str], force: bool) -> None:
+def configure_kubectl(config: str, env: Optional[str]) -> None:
     """
     Configure kubectl so you can connect to the cluster
 
@@ -40,7 +33,6 @@ def configure_kubectl(config: str, env: Optional[str], force: bool) -> None:
         event_properties={"org_name": layer.org_name, "layer_name": layer.name},
     )
     layer.verify_cloud_credentials()
-    if force:
-        purge_opta_kube_config(layer)
+    purge_opta_kube_config(layer)
     configure(layer)
     load_opta_kube_config_to_default(layer)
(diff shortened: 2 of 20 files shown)


modules/aws_eks/aws_eks.py
```diff
@@ -31,6 +32,7 @@ def post_delete(self, module_idx: int) -> None:
         self.cleanup_cloudwatch_log_group(region)
         self.cleanup_dangling_enis(region)
         self.cleanup_security_groups(region)
+        purge_opta_kube_config(layer=self.layer)
```
bigbitbus, line 35: We dont have a post_delete in the GCP GKE processor; I am wondering how the kubectl purge will work there?

bigbitbus (approved):
Looks good.

You may also want to update the documentation with the new command option.

modules/aws_eks/aws_eks.py
RemyDeWolf: move to `K8sServiceModuleProcessor.post_delete` to work for all clouds?

opta/commands/kubectl.py
```diff
@@ -16,7 +16,14 @@
...
     "-e", "--env", default=None, help="The env to use when loading the config file"
 )
-def configure_kubectl(config: str, env: Optional[str]) -> None:
+@click.option(
+    "-f",
+    "--force",
+    is_flag=True,
+    default=False,
+    help="Purge the existing cached kube config",
+)
```
RemyDeWolf, line 25: what about making this the default behavior so this option is not needed?

opta/core/kubernetes.py
```diff
@@ -66,24 +66,63 @@ def get_required_path_executables(cloud: str) -> FrozenSet[str]:
...
+        _aws_set_kube_config(layer)
     elif layer.cloud == "google":
-        _gcp_configure_kubectl(layer)
+        _gcp_set_kube_config(layer)
     elif layer.cloud == "azurerm":
-        _azure_configure_kubectl(layer)
+        _azure_set_kube_config(layer)
     elif layer.cloud == "local":
-        _local_configure_kubectl(layer)
+        _local_set_kube_config(layer)
```
RemyDeWolf, line 84: should we throw an error if not implemented (else case)?

opta/core/kubernetes.py
```diff
@@ -66,24 +66,63 @@ def get_required_path_executables(cloud: str) -> FrozenSet[str]:
...
 
 
-def _local_configure_kubectl(layer: "Layer") -> None:
+def purge_opta_kube_config(layer: "Layer") -> None:
+    """Delete the kubeconfig file created for a given layer, and also remove it from the default kubeconfig file"""
+    config_file_name = get_config_file_name(layer)
```
RemyDeWolf, line 88: suggestion: can this be renamed to kubeconfig_file_name (var and method)? to not be confused with the layer opta config file

opta/core/kubernetes.py
```diff
@@ -66,24 +66,63 @@ def get_required_path_executables(cloud: str) -> FrozenSet[str]:
...
+            for i, x in enumerate(default_kube_config[key])
+            if x["name"] == opta_value["name"]
+        ]
+        for index in sorted(current_indices, reverse=True):
+            del default_kube_config[key][index]
+
+    if default_kube_config["current-context"] == opta_config_context["name"]:
+        default_kube_config["current-context"] = ""
+    with open(default_kube_config_filename, "w") as f:
+        yaml.dump(default_kube_config, f)
```
RemyDeWolf, line 122: this block could be extracted out in it's own generic method `kubeconfig_unset(filename, user: Optional[str], context: Optional[str], cluster: Optional[str])`
such method could be unit tested without any opta objects

modules/aws_eks/aws_eks.py
juandiegopalomino: It's only needed for these two clouds, and they are not part of K8sServiceModuleProcessor

opta/core/kubernetes.py
juandiegopalomino: good point

opta/commands/kubectl.py
juandiegopalomino: good point

RemyDeWolf (approved):

PR #629, opened 2022-01-26
Fix for using relative paths

nsarupr:
# Description
Fix the relative path issue with k8s-manifest 

# Safety checklist

Tested manually.


juandiegopalomino (approved):

bigbitbus (approved):
tested, worked.

PR #634, opened 2022-01-27
Use stable k8s ingress version if detected

juandiegopalomino:
# Description
As our users start switching to the newer versions of kubernetes, we can intelligently determine if we can use the new stable version of the k8s ingress resource (and if we can we should).

@bigbitbus this is for AWS, we still need your pr for gcp

# Safety checklist

testing it right now manually

### DIFF
--- a/modules/aws_k8s_base/tf_module/opta-base/templates/_helpers.tpl
+++ b/modules/aws_k8s_base/tf_module/opta-base/templates/_helpers.tpl
@@ -59,3 +59,32 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- .Values.moduleName }}
 {{- end }}
 
+{{- define "ingress.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19.x" (include "prometheus.kubeVersion" .)) -}}
+      {{- print "networking.k8s.io/v1" -}}
+  {{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" -}}
+    {{- print "networking.k8s.io/v1beta1" -}}
+  {{- else -}}
+    {{- print "extensions/v1beta1" -}}
+  {{- end -}}
+{{- end -}}
+
+
+{{/*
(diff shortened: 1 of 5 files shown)


modules/aws_k8s_base/tf_module/opta-base/templates/_helpers.tpl
```diff
@@ -59,3 +59,32 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 {{- .Values.moduleName }}
 {{- end }}
 
+{{- define "ingress.apiVersion" -}}
+  {{- if and (.Capabilities.APIVersions.Has "networking.k8s.io/v1") (semverCompare ">= 1.19.x" (include "prometheus.kubeVersion" .)) -}}
```
NitinAgg, line 63: prometheus?

modules/aws_k8s_base/tf_module/opta-base/templates/_helpers.tpl
```diff
@@ -59,3 +59,32 @@ app.kubernetes.io/instance: {{ .Release.Name }}
...
+Return if ingress supports ingressClassName.
+*/}}
+{{- define "ingress.supportsIngressClassName" -}}
+  {{- or (eq (include "ingress.isStable" .) "true") (and (eq (include "ingress.apiVersion" .) "networking.k8s.io/v1beta1") (semverCompare ">= 1.18.x" (include "prometheus.kubeVersion" .))) -}}
+{{- end -}}
+{{/*
+Return if ingress supports pathType.
+*/}}
+{{- define "ingress.supportsPathType" -}}
+  {{- or (eq (include "ingress.isStable" .) "true") (and (eq (include "ingress.apiVersion" .) "networking.k8s.io/v1beta1") (semverCompare ">= 1.18.x" (include "prometheus.kubeVersion" .))) -}}
```
NitinAgg, line 89: prometheus is mentioned here again

bigbitbus:
As discussed, we have this sort of functionality via the other PR; lets park this and revisit when we have to upgrade the other deprecating APIs. Thanks @juandiegopalomino for this.

PR #653, opened 2022-02-01
Ghost example

NitinAgg:
# Description

# Safety checklist

YOUR_ANSWER


examples/ghost/README.md
```diff
@@ -0,0 +1,44 @@
...
+* Update `delegated` field to `true` in the `ghost-aws.yaml` file
+* Run `opta apply -c ghost-aws.yaml` again to generate the TLS certificate
+
+Your domain should now be pointing to the Ghost deployment with secure TLS
+
+* The admin dashboard would be available at https://{your-domain}/dashboard/ghost
+
+
+![Alt text](ghost-dashboard.png?raw=true "What it should look like")
+
```
juandiegopalomino, line 41: instructions on email setup?

examples/ghost/README.md
```diff
@@ -0,0 +1,44 @@
+# What is this?
+
+This is an example [Opta](https://github.com/run-x/opta) configuration file to deploy [Ghost](https://github.com/TryGhost/Ghost) on AWS.
+
+You can also use this button to fill in the variable fields in the config.
+
+[![Deploy](https://raw.githubusercontent.com/run-x/opta/main/assets/deploy-to-aws-button.svg)](https://app.runx.dev/deploy-with-aws?url=https%3A%2F%2Fgithub.com%2Frun-x%2Fopta%2Fblob%2Fmain%2Fexamples%2Fghost%2Fghost-aws.yaml&name=Ghost)
+
+# What does this do?
+It deploys a horizontally scalable Ghost deployment on EKS in AWS. It uses AWS managed the RDS mysql instance. It also sets up various other resources like VPCs, subnets, load balancers etc.
```
RemyDeWolf, line 10: It uses AWS managed _for_ the RDS mysql instance

examples/ghost/ghost-aws.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - type: k8s-base
+  - type: mysql
+    name: mysql
+    safety: false
+  - type: k8s-service
+    env_vars:
+      - name: database__client
+        value: "mysql"
+      - name: url
+        value: "https://${{module.dns.domain}}/dashboard"
```
RemyDeWolf, line 22: would that work even if the user only does the section `# Steps to deploy`?

examples/ghost/ghost-aws.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - name: NODE_ENV
+        value: "production node current/index.js"
+      - name: mail__transport
+        value: "SMTP"
+      - name: mail__options__service
+        value: "Gmail"
+      - name: mail__options__auth__user
+        value: ""
+    secrets:
+      - mail__options__auth__pass
```
RemyDeWolf, line 32: secrets are no longer required, if this needs to be set by the user they would run `opta secret update` for it.

examples/ghost/README.md
```diff
@@ -0,0 +1,44 @@
+# What is this?
```
RemyDeWolf, line 1: don't forget to list this example in the example landing page as well

examples/ghost/ghost-aws.yaml
NitinAgg: Yes, but would still leave this in as it clarifies what secret needs to be passed in.

examples/ghost/ghost-aws.yaml
NitinAgg: Yup this works. they can still write any random domain in the domain section and not delegate. and then just use the load balancer url

examples/ghost/ghost-aws.yaml
NitinAgg: Yup. So they only need this if they want to use the email feature which is not necessary. And I just documented the steps to setup the email

RemyDeWolf: cool, I'll double check on this as well - updating a secret with a env file containing a service

RemyDeWolf (approved):

PR #655, opened 2022-02-01
Redacting all sensitive information out of non stdout logs

juandiegopalomino:
# Description
Currently only removing the gcp project name or aws account id as opta doesn't show any other secrets.

# Safety checklist

Ran opta apply manually with an injected error for both gcp and aws and, unlike before, now the crash report does not mention my account id/project at all.


NitinAgg (commented):
Can you add some tests. Thanks

opta/utils/__init__.py
```diff
@@ -21,6 +21,40 @@
...
+            r"([\"\: \'\n]?)project([\"\: \'\n]?): ([\"\: \'\n]?)[a-zA-Z0-9_\-]+([\"\: \'\n]?)",
+            "\g<1>project\g<2>: \g<3>REDACTED\g<4>",
+        ),
+        (r"projects/[a-zA-Z0-9_\-]+/", "projects/REDACTED/"),
+        (
+            r"@[a-zA-Z0-9_\-]+.iam.gserviceaccount.com",
+            "@REDACTED.iam.gserviceaccount.com",
+        ),
+        (r"[a-zA-Z0-9_\-]+.svc.id.goog", "REDACTED.svc.id.goog",),
+        (r"gcr.io/[a-zA-Z0-9_\-]+", "gcr.io/REDACTED"),
```
rabbitfang, line 40:
```suggestion
        (
            r"@[a-zA-Z0-9_\-]+\.iam\.gserviceaccount\.com",
            "@REDACTED.iam.gserviceaccount.com",
        ),
        (r"[a-zA-Z0-9_\-]+\.svc\.id\.goog", "REDACTED.svc.id.goog",),
        (r"gcr\.io/[a-zA-Z0-9_\-]+", "gcr.io/REDACTED"),
```

Treat the `.` in domain names as literals.

rabbitfang (changes requested):
Tests should be added to this so we can verify that the redaction is happening as expected.

opta/crash_reporter.py
```diff
@@ -79,5 +79,5 @@ def emit(self, record: LogRecord) -> None:
 CURRENT_CRASH_REPORTER = CrashReporter()
 crash_report_handler = ListHandler(CURRENT_CRASH_REPORTER.log_lines)
 crash_report_handler.setLevel(DEBUG)
-crash_report_handler.setFormatter(LogFormatMultiplexer())
+crash_report_handler.setFormatter(SensitiveFormatter())
```
legalosLOTR, line 82: Did you mean to replace LogFormatMultiplexer with SensitiveFormatter? Looks like you've already added SensitiveFormatter to the LogFormatMultiplexer class itself.

juandiegopalomino: Yes, this is intended

juandiegopalomino: We weren't multiplexing anyting there so it's fine

legalosLOTR: cool!

NitinAgg (approved):

rabbitfang (approved):

PR #656, opened 2022-02-02
Remove aws_s3_bucket_public_access_block from Buckets

nsarupr:
# Description
This change allows users to add public access to objects and buckets.

# Safety checklist


juandiegopalomino (approved):

bigbitbus (commented):
@nsarupr did you check what happens to pre-existing opta buckets from old deploys? Want to make sure they dont get clobbered/recreated etc.

PR #659, opened 2022-02-03
OPTA DEPLOY User optimization

nsarupr:
# Description
Allow users to use `opta deploy` with custom images (other than AUTO).
This also makes `-i/--image` optional, and validates based on the `image` value present in the configuration.
If the image is `AUTO` it will require you to provide a user defined docker image and push the image to the Container Registry.
If the image is `xyz`, it will just pull that image directly and run it.

# Safety checklist

### DIFF
--- a/opta/commands/deploy.py
+++ b/opta/commands/deploy.py
@@ -18,7 +18,7 @@ from opta.utils import check_opta_file_exists, fmt_msg, logger
 
 @click.command()
 @click.option(
-    "-i", "--image", required=True, help="Your local image in the for myimage:tag"
+    "-i", "--image", help="Your local image in the for myimage:tag", default=None
 )
 @click.option("-c", "--config", default="opta.yaml", help="Opta config file.")
 @click.option(
@@ -118,6 +118,7 @@ def deploy(
         amplitude_client.DEPLOY_EVENT,
         event_properties={"org_name": layer.org_name, "layer_name": layer.name},
     )
+    custom_image_name = __check_layer_and_image(layer, image)
     layer.verify_cloud_credentials()
     layer.validate_required_path_dependencies()
     if Terraform.download_state(layer):
@@ -129,22 +130,25 @@ def deploy(
         outputs = Terraform.get_outputs(layer)
     except MissingState:
         outputs = {}
-    if "docker_repo_url" not in outputs or outputs["docker_repo_url"] == "":
-        logger.info(
-            "Did not find docker repository in state, so applying once to create it before deployment"
-        )
-        _apply(
-            config=config,
-            env=env,
-            refresh=False,
-            image_tag=None,
-            test=False,
-            local=local,
-            auto_approve=auto_approve,
-            stdout_logs=False,
-            detailed_plan=detailed_plan,
-        )
-    image_digest, image_tag = _push(image=image, config=config, env=env, tag=tag)
+
+    image_digest, image_tag = (None, None)
+    if custom_image_name == "AUTO":
+        if "docker_repo_url" not in outputs or outputs["docker_repo_url"] == "":
+            logger.info(
+                "Did not find docker repository in state, so applying once to create it before deployment"
+            )
+            _apply(
+                config=config,
+                env=env,
+                refresh=False,
+                image_tag=None,
+                test=False,
+                local=local,
+                auto_approve=auto_approve,
+                stdout_logs=False,
+                detailed_plan=detailed_plan,
+            )
+        image_digest, image_tag = _push(image=image, config=config, env=env, tag=tag)
     _apply(
         config=config,
         env=env,
@@ -156,3 +160,13 @@ def deploy(
         image_digest=image_digest,
         detailed_plan=detailed_plan,
     )
+
+
+def __check_layer_and_image(layer: "Layer", image: str) -> str:
+    k8s_module = layer.get_module_by_type("k8s-service")
+    image_name = k8s_module[0].data.get("image")
+    if image_name == "AUTO" and image is None:
+        raise UserErrors("An image should be passed when using `image` as AUTO in configuration")
+    if image_name != "AUTO" and image is not None:
+        raise UserErrors(f"Do not pass any image. Image {image_name} already present in configuration.")
+    return image_name


opta/commands/deploy.py
```diff
@@ -129,22 +130,25 @@ def deploy(
...
-            test=False,
-            local=local,
-            auto_approve=auto_approve,
-            stdout_logs=False,
-            detailed_plan=detailed_plan,
-        )
-    image_digest, image_tag = _push(image=image, config=config, env=env, tag=tag)
+
+    image_digest, image_tag = (None, None)
+    if custom_image_name == "AUTO":
```
juandiegopalomino, line 135: can be lower case

opta/commands/deploy.py
```diff
@@ -156,3 +160,17 @@ def deploy(
         image_digest=image_digest,
         detailed_plan=detailed_plan,
     )
+
+
+def __check_layer_and_image(layer: "Layer", image: str) -> str:
+    k8s_module = layer.get_module_by_type("k8s-service")
+    image_name = k8s_module[0].data.get("image")
+    if image_name == "AUTO" and image is None:
+        raise UserErrors(
```
juandiegopalomino, line 169: Don't print an error-- this will just remove the existing pods running, and is useful when doing the first deploy

opta/commands/deploy.py
```diff
@@ -156,3 +160,17 @@ def deploy(
...
+            "An image should be passed when using `image` as AUTO in configuration"
+        )
+    if image_name != "AUTO" and image is not None:
+        raise UserErrors(
+            f"Do not pass any image. Image {image_name} already present in configuration."
```
RemyDeWolf, line 174: you can also add ` or set the image value to AUTO`

opta/commands/deploy.py
```diff
@@ -156,3 +160,17 @@ def deploy(
         image_digest=image_digest,
         detailed_plan=detailed_plan,
     )
+
+
+def __check_layer_and_image(layer: "Layer", image: str) -> str:
```
RemyDeWolf, line 165: I would suggest to keep this super simple by removing the layer, something like
```
image_name, is_custom = __check_image(image_from_config: str, image_from_option: str)
# these names make the code easier to read
```
we would only compute `is_custom` in this function and use the boolean var then

opta/commands/deploy.py
```diff
@@ -18,7 +18,7 @@
 
 @click.command()
 @click.option(
-    "-i", "--image", required=True, help="Your local image in the for myimage:tag"
+    "-i", "--image", help="Your local image in the for myimage:tag", default=None
```
RemyDeWolf, line 21: now that this is optional let's also update the doc to reflect this and add an example https://github.com/run-x/opta/blob/d8eb1a3c78f8c58f267ff68af66b97b8b7ca6677/opta/commands/deploy.py#L63-L76

opta/commands/deploy.py
nsarupr: More so...let me make this case insensitive.

opta/commands/deploy.py
nsarupr: Sure, will do.

opta/commands/deploy.py
```diff
@@ -118,6 +119,7 @@ def deploy(
         amplitude_client.DEPLOY_EVENT,
         event_properties={"org_name": layer.org_name, "layer_name": layer.name},
     )
+    _, is_auto = __check_layer_and_image(layer, image)
```
RemyDeWolf, line 122: I didn't realize we would not need `image_name` then we can just simply have this function returning only `is_auto`

opta/commands/deploy.py
```diff
@@ -129,22 +131,26 @@ def deploy(
...
-            image_tag=None,
-            test=False,
-            local=local,
-            auto_approve=auto_approve,
-            stdout_logs=False,
-            detailed_plan=detailed_plan,
-        )
-    image_digest, image_tag = _push(image=image, config=config, env=env, tag=tag)
+
+    image_digest, image_tag = (None, None)
```
RemyDeWolf, line 135: something seems off with `image_tag` it's always None when calling apply?
can you double check?
and cover by unit test if indeed there was an error.

if we don't need it then we don't need this var

tests/commands/test_deploy.py
```diff
@@ -13,47 +13,6 @@ def mock_is_service_config(module_mocker: MockFixture) -> None:
     module_mocker.patch("opta.commands.deploy.is_service_config", return_value=True)
 
 
-def test_check_layer_and_image_auto_image_no_image_flag(mocker: MockFixture) -> None:
```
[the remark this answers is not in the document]
nsarupr: Removed it.

opta/commands/deploy.py
```diff
@@ -58,19 +58,20 @@ def deploy(
     detailed_plan: bool,
     local: Optional[bool],
 ) -> None:
-    """Deploy a local image to Kubernetes
+    """Deploys an image to Kubernetes
 
-    1. Push the image to the private container registry (ECR, GCR, ACR)
+    - Pushes the local image to private container registry (ECR, GCR, ACR), iff configuration contains `image: AUTO`,
```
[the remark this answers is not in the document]
nsarupr: Kind if meant, `if and only if`

opta/commands/deploy.py
nsarupr: Right now, every time the push occurs, we are passing in the image digest which is different every time.

opta/commands/deploy.py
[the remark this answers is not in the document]
RemyDeWolf: ah, I didn't know about `iff`, but that makes me think that other people will not know about it either

opta/commands/deploy.py
nsarupr: Changed.

opta/commands/deploy.py
RemyDeWolf: cool, something we will revisit later

RemyDeWolf (approved):

PR #688, opened 2022-02-10
Bug/add ec2 autoscaler permission

bigbitbus:
# Description
Missing permission added to IAM role; this probably broke when we upgraded to EKS 1.21, now fixed via the extra permission added here.

https://run-x.atlassian.net/browse/RUNX-1162

# Safety checklist

Manual test


bigbitbus:
Closing this and sending another one - commits got mixed up

PR #704, opened 2022-02-11
Generate terraform files given Opta file

RemyDeWolf:
# Description
Implement #672 

See the 5min video for more context: https://loom.com/share/8e3bdefef98c4d69bc67e4e051c323db

```
opta generate-terraform --config env.yaml
Terraform files generated, check ./generated-terraform/staging.md for documentation

ls ./generated-terraform
data.tf.json              module-k8sbase.tf.json    modules                   provider.tf.json          terraform.tf.json
module-base.tf.json       module-k8scluster.tf.json output.tf.json            staging.md

```

# Implementation notes
- This command doesn't need any cloud provider to work. 
- Some modules like k8-base connect to k8s in their process method, a stateless flag was introduced so such logic can be disabled at runtime.
- The terraform modules have to be created in a specific order and would vary depending on their stack, a readme is generated for the user to document how to proceed. A small markdown utility was implemented with this PR.
- Since we can't expect any infra to exist (such a S3 bucket), the terraform storage is changed to local. And it will be up to the user to change it accordingly. But for a user that would like to switch from opta to terraform, they might want to keep the existing bucket, I will check if we can have an option for this.
- opta currently generate one file  main.tf.json  which contains all the configuration for providers, data, modules...  If a user would generate terraform for an environment and a service, these files would have duplication that would create terraform errors - ex: terraform can only have one provider named aws. To solve this issue, this file was split into smaller files (provider.tf.json, data.tf.json) and the json data will be merged if a user runs export on an existing directory. This allows the user to run generate-terraform for many opta configuration files.
- The command uses a temporary directory and only copy the generated files to the target directory if it succeed.

# Safety checklist

Tested on AWS with the getting started example, which includes the modules: base, k8s-cluster, k8s-base, k8s-service

# Next steps
- test with DB modules
- test with GCP
- support remote backend (right now it's only local)


opta/amplitude.py
```diff
@@ -32,6 +32,7 @@ class AmplitudeClient:
     VIEW_OUTPUT_EVENT = "VIEW_OUTPUT"
     FORCE_UNLOCK_EVENT = "FORCE_UNLOCK"
     FINISH_GEN_EVENT = "FINISH_GEN"
+    GEN_TERRAFORM = "GEN_TERRAFORM"
```
juandiegopalomino, line 35: Add a FINISH_GEN_TERRAFORM_EVENT, and use it like FINISH_GEN_EVENT so that we can count how many successes/failures there are

juandiegopalomino (approved):
amazing! just one comment

opta/commands/generate_terraform.py
```diff
@@ -0,0 +1,304 @@
...
+
+    Examples:
+
+    opta generate-terraform -c my-config.yaml
+
+    opta generate-terraform -c my-config.yaml -d ./tf
+    """
+
+    if directory.strip() == "":
+        raise click.UsageError("--directory can't be empty")
```
bigbitbus, line 80: We may want to add a check that the directory is not already populated with an older run of generate terraform; otherwise TF files may get mixed up...(e.g. I generate terraform for a staging env, then for a product env, and then all the files are mixed up in this directory.)

RemyDeWolf: good point, I added a confirmation check for this, see `The output directory will be updated.\n Do you approve?` below

opta/commands/generate_terraform.py
```diff
@@ -0,0 +1,304 @@
...
+            logger.debug(f"Copying helm charts from {service_helm_dir} to {target_dir}")
+            shutil.copytree(service_helm_dir, target_dir, dirs_exist_ok=True)
+
+        # copy module directories and update the module path to point to local directory
+        for module in layer.modules:
+            src_path = module.module_dir_path
+            rel_path = "./" + src_path[src_path.index("modules/") :]
+            abs_path = os.path.join(tmp_dir, rel_path)
+            logger.debug(f"Copying module from {module.aliased_type} to {abs_path}")
+            shutil.copytree(src_path, abs_path, dirs_exist_ok=True)
```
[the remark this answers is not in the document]
RemyDeWolf: right, it will only copy the `tf_module` I added a comment in the code as well to clarify

opta/commands/generate_terraform.py
```diff
@@ -0,0 +1,304 @@
...
+                _clean_folder(output_dir)
+            else:
+                logger.info(
+                    f"Output directory {directory} already exists, adding new files to it"
+                )
+
+        logger.debug(f"Copy {tmp_dir} to {output_dir}")
+        shutil.copytree(tmp_dir, output_dir, dirs_exist_ok=True)
+        logger.info("Terraform files generated successfully.")
+        if readme_file:
```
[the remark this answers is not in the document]
RemyDeWolf: really good point.
Here is what I've done:
- a module can be marked as `is_exportable: false` by default it would be true
- if generate terraform includes a unsupported module:
  - command output will be: Terraform files partially generated, a few modules are not supported:
  - in the generated readme, the  unsupported modules will be listed
  - if we want to include special documentation as well, we can define a file called `export.md` in the module directory, this file will be added to the generated readme. I added some examples in this PR with datadog and aws-dns

as we ramp up the testing, we will mark deal with the unsupported module:
- either mark them as `is_exportable: false`
- or refactor them to be exportable

RemyDeWolf (commented):
@bigbitbus @juandiegopalomino  thanks for the review, addressed them, it's ready for another look.

opta/commands/generate_terraform.py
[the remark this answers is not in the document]
bigbitbus: thanks

opta/commands/generate_terraform.py
bigbitbus: thanks

PR #707, opened 2022-02-11
Added field to let users configure aurora backup retention period

juandiegopalomino:
# Description
I'm calling it backup_retention_days

# Safety checklist

n/a

### DIFF
--- a/modules/aws_mysql/aws-mysql.json
+++ b/modules/aws_mysql/aws-mysql.json
@@ -15,7 +15,7 @@
       "default": "5.7.mysql_aurora.2.04.2"
     },
     "backup_retention_days": {
-      "type": "boolean",
+      "type": "integer",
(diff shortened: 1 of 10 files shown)


modules/aws_mysql/aws-mysql.json
```diff
@@ -14,6 +14,11 @@
       "description": "The version of the database to use.",
       "default": "5.7.mysql_aurora.2.04.2"
     },
+    "backup_retention_days": {
+      "type": "boolean",
+      "description": "How many days to keep the backup retention",
+      "default": 7
```
bigbitbus, line 20: Boolean or number?

bigbitbus (approved):
Just a minor comment, lgtm

NitinAgg (approved):

PR #708, opened 2022-02-12
Deleting EKS cloudwatch log group if pre-existing

juandiegopalomino:
# Description
Cloudwatch log groups have this nasty habit of reviving themselves if there's ever one more log sent to them. This is why even though terraform deletes  them on opta destroy and we try to delete it again on post delete, they still stuck around. The consensus was to catch the log groups if they stuck around after a destroy and delete them there.

# Safety checklist

YOUR_ANSWER


modules/aws_eks/aws_eks.py
```diff
@@ -94,6 +95,29 @@ def cleanup_cloudwatch_log_group(self, region: str) -> None:
                 "wanna check this out."
             )
 
+    def delete_preexisting_cloudwatch_log_group(self, region: str) -> None:
+        """If there is a pre-existing cloudwatch log group (which can happen if the opta module was destroyed, but the
+        log group was revived because of dalyed final log messages), then delete the log group"""
```
RemyDeWolf, line 100: _delayed_

RemyDeWolf (commented):
I thought that the consensus was to delete before creating the cluster not after right?

modules/aws_eks/aws_eks.py
```diff
@@ -192,4 +200,7 @@ def process(self, module_idx: int) -> None:
             "kms_account_key_arn"
         ] = f"${{{{module.{aws_base_module.name}.kms_account_key_arn}}}}"
         self.module.data["vpc_id"] = f"${{{{module.{aws_base_module.name}.vpc_id}}}}"
+
+        providers = self.layer.gen_providers(0)
+        region = providers["provider"]["aws"]["region"]
```
RemyDeWolf, line 205: why is this change needed? I'm not following this

RemyDeWolf (approved):

PR #713, opened 2022-02-15
Update package.yml

nsarupr:
# Description

# Safety checklist


NitinAgg (approved):

rabbitfang (commented):
We should name the binaries with the system architecture that it applies to (e.g. for linux, amd64 or arm)

PR #716, opened 2022-02-15
Feature/unique tf state bucket name

bigbitbus:
# Description
Making state bucket names unique. 

For pre-existing deployments, the same old bucket is used.

For new deployments, the state_storage bucket name has a 4 hexadecimal character suffixed; derived from hash of region/provider_id | project (aws | google); Azure is also addressed in the PR.

Uses functools caching to avoid multiple API calls when checking bucket existence (thanks @NitinAgg )

# Safety checklist

Create/destroy on new and pre-existing aws/gcp environments.

### DIFF
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -374,7 +374,6 @@ class AWS(CloudClient):
         return True
 
 
-
 # AWS Resource ARNs can be one of the following 3 formats:
 # 1). arn:partition:service:region:account-id:resource-id
 # 2). arn:partition:service:region:account-id:resource-type/resource-id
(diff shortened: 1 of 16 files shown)


opta/core/aws.py
```diff
@@ -362,6 +363,16 @@ def parse_arn(arn: str) -> AwsArn:
             result["resource_type"], result["resource"] = result["resource"].split(":", 1)
         return result
 
+    @staticmethod
+    def bucket_exists(bucket_name: str, region: str) -> bool:
```
juandiegopalomino, line 367: could we have them all be static or not static?

juandiegopalomino: actually nvmnd, it's a pointless conformity

juandiegopalomino (approved):
lgtm

PR #717, opened 2022-02-16
generate-terraform: support remote backend

RemyDeWolf:
follow up implementation for #672
# Description
For a user that wants to keep Opta but just want to check their terraform plan, supporting the remote backend will be needed.

```
opta generate-terraform --backend remote -c opta.yaml
Terraform files generated successfully.

cat ./generated-terraform/terraform.tf.json
{
    ....
    "backend": {
      "gcs": {
        "bucket": "opta-tf-state-runx-staging",
        "prefix": "staging"
      }
    }
  }
}
```

Also include feedback from Nilesh
- rename `--replace` to `--delete`
- specify that the generated command includes default values
- provide instructions to connect to k8s

# Safety checklist

Manual test + added unit tests.

### DIFF
--- a/opta/commands/generate_terraform.py
+++ b/opta/commands/generate_terraform.py
@@ -204,7 +204,7 @@ def generate_terraform(
             layer, execution_plan, tmp_dir, readme_format, opta_cmd, backend
         )
 
-        # if everything was successfull, copy tmp dir to target dir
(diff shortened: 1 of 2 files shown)


nsarupr (approved):
Looks good to me.

PR #718, opened 2022-02-17
CircleCI test for Terraform generated State files for AWS.

nsarupr:
# Description
Using the State files generated using the command `opta generate-terraform`

# Safety checklist

Successful CircleCI test: **https://app.circleci.com/pipelines/github/run-x/opta/1431/workflows/af39af94-854e-4b8f-b487-8d80a1ecb344**

### DIFF
--- a/.circleci/ci-tests/terraform-generated-aws/service-http.yaml
+++ b/.circleci/ci-tests/terraform-generated-aws/service-http.yaml
@@ -1,0 +1,20 @@
+environments:
+  - name: env-tg
+    path: "environment.yaml"
+    variables:
+      max_containers: 2
+name: http-service
+modules:
+  - name: app
+    type: k8s-service
+    image: kennethreitz/httpbin
+    public_uri: "/"
+    min_containers: 1
+    max_containers: "{vars.max_containers}"
+    liveness_probe_path: "/get"
+    readiness_probe_path: "/get"
+    sticky_session: true
+    sticky_session_max_age: 20
+    keep_path_prefix: true
+    port:
+      http: 80
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -368,31 +368,24 @@ jobs:
     executor: ubuntu-machine
     steps:
       - install-opta-dependencies
-      # - checkout-opta
       - attach_workspace:
           at: ~/
       - run:
-          name: "Test Terraform Generated Environment"
+          name: "Create Terraform Generated Environment"
           command: |
             cd $HOME/generated-tf-files-aws
-            terraform init          
-      # - run:
-      #     name: "Create Terraform Generated Environment"
-      #     command: |
-      #       cd $HOME/generated-tf-files-aws
-      #       terraform init
-      #       terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.base
-      #       terraform apply -compact-warnings -auto-approve tf.plan
-      #       terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.k8scluster
-      #       terraform apply -compact-warnings -auto-approve tf.plan
-      #       terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.k8sbase
-      #       terraform apply -compact-warnings -auto-approve tf.plan
+            terraform init
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.base
+            terraform apply -compact-warnings -auto-approve tf.plan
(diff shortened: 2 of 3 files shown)


.circleci/ci-tests/terraform-generated-aws/service-http.yaml
```diff
@@ -0,0 +1,20 @@
+environments:
+  - name: env-tg
+    path: "environment.yaml"
+    variables:
+      max_containers: 2
+name: http-service
+modules:
+  - name: app
+    type: k8s-service
+    image: kennethreitz/httpbin
```
RemyDeWolf, line 10: note: getting started use `ghcr.io/run-x/opta-examples/hello-app:main` which is very similar but not subject to `[docker hub rate limit](https://docs.docker.com/docker-hub/download-rate-limit/)`, which we might eventually hit.

.circleci/config.yml
```diff
@@ -329,6 +333,115 @@ jobs:
...
+      - attach_workspace:
+          at: ~/
+      - run:
+          name: "Create Terraform Generated Environment"
+          command: |
+            cd $HOME/generated-tf-files-aws
+            terraform init
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.base
+            terraform apply -compact-warnings -auto-approve tf.plan
+            cat ./tfstate/awsenv-tg.tfstate
```
RemyDeWolf, line 373: why do we need the cat here?
it will make the log too large

.circleci/config.yml
```diff
@@ -329,6 +333,115 @@ jobs:
...
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.k8scluster
+            terraform apply -compact-warnings -auto-approve tf.plan
+            cat ./tfstate/awsenv-tg.tfstate
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.k8sbase
+            terraform apply -compact-warnings -auto-approve tf.plan
+            cat ./tfstate/awsenv-tg.tfstate
+      - persist_to_workspace:
```
RemyDeWolf, line 380: 👍

.circleci/config.yml
```diff
@@ -329,6 +333,115 @@ jobs:
...
+          command: |
+            cd $HOME/generated-tf-files-aws
+            terraform init
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.base -target=module.k8scluster -target=module.k8sbase -destroy
+            terraform apply -compact-warnings -auto-approve tf.plan
+      - persist_to_workspace:
+          root: ~/
+          paths:
+            - generated-tf-files-aws/tfstate/awsenv-tg.tfstate
+
```
RemyDeWolf, line 444: we don't need to persist after destroy right?

.circleci/ci-tests/terraform-generated-aws/service-http.yaml
nsarupr: I don't understand what you mean by that. 
Are you in favour of using the `kennethreitz/httpbin` image or our own Hello world app.
As for the pull rate limit, it's 200 per 6 hrs, so that shouldn't be of any worry to us.

.circleci/config.yml
nsarupr: This is a temporary solution so that if there is a fail, we would have the state files if need arises to pull the state in a different machine, since currently it would be stored locally, that that would just go away since the CI machines are ephemeral.

.circleci/config.yml
nsarupr: Yeah we don't...let me remove that...

.circleci/ci-tests/terraform-generated-aws/service-http.yaml
RemyDeWolf: yeah not a big deal, but if we use this image in all our tests and we run test for every merge we might eventually hit this limit.

.circleci/config.yml
RemyDeWolf: you mean to troubleshoot? 
we can use artifacts for that https://circleci.com/docs/2.0/artifacts/

.circleci/ci-tests/terraform-generated-aws/service-http.yaml
nsarupr: I am using this image because this is a standard image we use for testing all things internal, hence chose to use this.

RemyDeWolf: this is fine for this PR, but may be we should start using the new image with new tests, to use the same image that is used in our docs (dogfooding)

RemyDeWolf (approved):

PR #726, opened 2022-02-22
opta config view --cloud (aws/google)

nsarupr:
# Description
Feature to view all the opta related configurations present in the cloud.
The feature currently supports AWS and GCP.

# Safety checklist

Tested Manually.


opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
+from typing import Optional
```
RemyDeWolf, line 1: we need unit test to cover this class similar to the other commands.
you can look at `test_generate_terraform` it's the latest we added

opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
...
+
+from opta.core.aws import AWS
+from opta.core.azure import Azure
+from opta.core.cloud_client import CloudClient
+from opta.core.gcp import GCP
+from opta.utils import logger
+
+
+@click.group(cls=DYMGroup)
+def config() -> None:
```
RemyDeWolf, line 14: add the user documentation here, ex opta/commands/secret.py

opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
...
+    pass
+
+
+cloud_option = click.option(
+    "-c",
```
RemyDeWolf, line 19: let's not use `-c` because we use it everywhere else for `--config` but this has a different meaning here
it's fine to only support `--cloud` not shortcut

opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
...
+)
+
+name_option = click.option(
+    "-n", "--name", help="The name of the configuration", required=False
+)
+
+environment_option = click.option(
+    "-e", "--environment", help="The name of the environment", required=False
+)
+
```
RemyDeWolf, line 33: let's stick to MVP, `opta config view --cloud aws/google/azurerm` 
I don't think we really had come up to a point where we thought the other options enough to implement them.

opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
...
+@config.command(name="list")
+@cloud_option
+@environment_option
+def list_command(cloud: str, environment: Optional[str]) -> None:
+    if environment:
+        logger.info("Environment option is a feature in progress.")
+    cloud_client: CloudClient = __get_cloud_client(cloud)
+    config_map = cloud_client.get_config_map()
+    if config_map:
+        logger.info(config_map)
```
RemyDeWolf, line 44: again let's stick to `view`

opta/commands/config.py
```diff
@@ -0,0 +1,73 @@
...
+        for bucket, detailed_configs in detailed_config_map.items():
+            for config_name, actual_config in detailed_configs.items():
+                logger.info(
+                    f"Bucket Name: {bucket}\nConfig Name: {config_name}\nActual Config:\n{actual_config}"
+                )
+
+
+def __get_cloud_client(cloud: str) -> CloudClient:
+    cloud_client: CloudClient
+    if cloud == "aws":
```
RemyDeWolf, line 66: suppor ignore case for these checks?

opta/core/cloud_client.py
```diff
@@ -28,3 +28,21 @@ def delete_remote_state(self) -> None:
...
+
+    @classmethod
+    def get_config_map(cls) -> Dict[str, List[str]]:
+        pass
+
+    @classmethod
+    def get_detailed_config_map(
+        cls, environment: Optional[str] = None
+    ) -> Dict[str, Dict[str, str]]:
+        pass
```
RemyDeWolf, line 48: I might be missing something but we should only need **one method to implement per cloud** (to get the content of the files in the remote bucket opta_config).
Every other operation, such as formatting, data manipulation is not cloud specific and could use the same implementation for each cloud.

```
opta config view --aws
```
will do:
1 - call aws.get_config()
2 - (cloud specific) returns content of `opta_config/awsenv-ci` 
3 - render the content to look as opta config files
4 - possible future improvement: do some filtering

only 2 would need to be implemented per cloud.

RemyDeWolf (commented):
couple of comments, let's stick to MVP, the cloud specific implementation is very costly to maintain, whenever possible we should avoid it. Let me know what you think about my comment of only having one method to implement and do the processing operations in common.

opta/core/cloud_client.py
```diff
@@ -28,3 +28,7 @@ def delete_remote_state(self) -> None:
     @abstractmethod
     def get_terraform_lock_id(self) -> str:
         raise NotImplementedError()
+
+    @classmethod
+    def get_detailed_config_map(cls) -> Dict[str, Dict[str, str]]:
```
RemyDeWolf, line 33: since we already have
```
    @abstractmethod
    def get_remote_config(self) -> Optional["StructuredConfig"]:
        pass
```

let's call this new method using a similar name
```
    @abstractmethod
    def list_remote_configs(self) -> Dict[str, "StructuredConfig"]]:
       #TODO add docstring
        pass
```
list_ is a better prefix some similar name to `s3.list_objects`

We should be able to reuse the implementation done `get_remote_config` by creating a new private method   `def _get_remote_config(layer_name str) that could be called by both `get_remote_config` and `list_remote_configs` . That would avoid duplicating some similar code. 


it would be up to the view command call `["original_spec"]` when displaying the info to the user

opta/core/aws.py
```diff
@@ -167,6 +167,29 @@ def force_delete_terraform_lock_id(self) -> None:
...
+                    detailed_configs[data["Key"][len(prefix) :]] = json.loads(
+                        config_object["Body"].read()
+                    )["original_spec"]
+                opta_config_detailed_map[aws_bucket] = detailed_configs
+        return opta_config_detailed_map
+
+    @classmethod
+    def get_bucket_list(cls) -> List[str]:
+        s3 = boto3.client("s3")
+        aws_bucket_data = s3.list_buckets().get("Buckets", [])
```
RemyDeWolf, line 190: this method can take a parameter called `Prefix` we should use it.
check if there is something similar for gcp

opta/commands/config.py
```diff
@@ -0,0 +1,49 @@
...
+    required=True,
+    type=click.Choice(["aws", "google", "azurerm"]),
+)
+
+
+@config.command()
+@cloud_option
+def view(cloud: str) -> None:
+    if cloud.lower() == "azurerm":
+        raise UserErrors("Currently AzureRM isn't supported.")
```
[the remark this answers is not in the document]
nsarupr: This one comes up when there is a user error raised.

opta/core/local.py
```diff
@@ -60,3 +60,9 @@ def delete_remote_state(self) -> None:
 
     def get_terraform_lock_id(self) -> str:
         return ""
+
+    def list_child_config_names(self) -> List[str]:
+        return []
+
+    def get_configuration_details(self, config_name: str) -> Dict[str, Any]:
+        return {}
```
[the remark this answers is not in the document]
nsarupr: I'll remove this.

opta/core/cloud_client.py
nsarupr: I'd rather we have these 2 separate functionality as of now...because they both are serving a different purpose. Also, adding `["original_spec"]` to the `view` isn't clean. If at any point we decide to change how we store config, we'll have to make changes to the command, which isn't required.
Also, `get_remote_config` serves 2 purposes,  one is checking the version of opta with which the commands were run, and another is getting details for the modules for the new terraform generation purpose.
As of now, what we can do is make the download part common for the both of them, as that's the only functionality which I see to be common among the both of them.

opta/core/aws.py
nsarupr: We can think about this, but maybe in an incremental stage. When we have a concrete thing on how we want the bucket names to look like. Once we have that fixed, we can add some sort of prefix. But till that time, let's use the logic to find the `opta_config` directory to figure out if the current bucket is an opta bucket or not.

opta/core/cloud_client.py
nsarupr: Also, we can extend this functionality at a later stage when we just don't need the actual configs but the lists of configurations or some other functionality.

opta/commands/config.py
```diff
@@ -0,0 +1,62 @@
...
+from opta.core.cloud_client import CloudClient
+from opta.core.gcp import GCP
+from opta.exceptions import UserErrors
+from opta.utils import logger
+
+
+@click.group(cls=DYMGroup)
+def config() -> None:
+    """
+    (Beta) Manage opta configurations.
```
RemyDeWolf, line 15: the terraform generator is in beta, because we have not yet tested all the modules, but this feature does not need to be tested for each module, do we need to be _beta_?

opta/commands/config.py
```diff
@@ -0,0 +1,62 @@
...
+    required=True,
+    type=click.Choice(["aws", "google", "azurerm"]),
+)
+
+
+@config.command()
+@cloud_option
+def view(cloud: str) -> None:
+    """
+    View the opta configuration file for a given cloud provider.
```
RemyDeWolf, line 32: for consistency no dot on the first line that is shown in the helper

opta/commands/config.py
```diff
@@ -0,0 +1,62 @@
...
+    Use the current cloud credentials to fetch the remote storage.
+
+    Examples:
+
+        - View opta configurations:
```
RemyDeWolf, line 37: we don't need this line, create extra ident and the command itself is self descriptive
check
`opta secret view -h `

opta/commands/config.py
```diff
@@ -0,0 +1,62 @@
...
+    type=click.Choice(["aws", "google", "azurerm"]),
+)
+
+
+@config.command()
+@cloud_option
+def view(cloud: str) -> None:
+    """
+    View the opta configuration file for a given cloud provider.
+    Use the current cloud credentials to fetch the remote storage.
```
RemyDeWolf, line 33: it's not remote _storage_ it's the remote opta config file.

opta/core/azure.py
```diff
@@ -170,3 +172,38 @@ def bucket_exists(self, bucket_name: str) -> bool:
             return True
         except ResourceNotFoundError:
             return False
+
+    @classmethod
+    def get_detailed_config_map(cls) -> Dict[str, Dict[str, str]]:
```
RemyDeWolf, line 177: every line of code that we add cost us in maintenance, so since the config command doesn't support azure, we should not keep the implementation.

opta/core/cloud_client.py
RemyDeWolf: > Also, get_remote_config serves 2 purposes, one is checking the version of opta with which the commands were run, and another is getting details for the modules for the new terraform generation purpose.

What is the part with checking the version, I don't see it in the impl https://github.com/run-x/opta/blob/main/opta/core/aws.py#L78-L90

I have an issue with this name `get_detailed_config_map` because it sounds like k8s ConfigMap. And in the existing code we already use the names `remote_configs` and `original_spec`.

What you can do also is:
```
    @abstractmethod
    def list_remote_configs(self) -> Dict[str, "StructuredConfig"]]:
       #TODO add docstring
        pass

   def list_original_spec(self) -> Dict[str, str]]:
       #TODO add docstring
        # implement using list_remote_configs  and return ["original_spec"]
```

The method `list_original_spec` would be implemented in one place.

opta/core/gcp.py
```diff
@@ -173,3 +173,22 @@ def bucket_exists(self, bucket_name: str) -> bool:
...
+        prefix = "opta_config/"
+        credentials, project_id = cls.get_credentials()
+        storage_client = storage.Client(project=project_id, credentials=credentials)
+        opta_config_detailed_map = {}
+        for bucket in storage_client.list_buckets():
+            detailed_config = {}
+            for response in storage_client.list_blobs(
+                bucket.name, prefix=prefix, delimiter="/"
+            ):
+                blob = storage.Blob(response.name, bucket)
```
RemyDeWolf, line 188: if you compare this new code and the existing code here, you can see the opportunity of creating a private method for loading a config file based on bucket and a file name.
https://github.com/run-x/opta/blob/498aa95bacb9d52a919ed13dd40a9581cffdbd76/opta/core/gcp.py#L83-L90

let me know if you need help with addressing this comment, but this is important that we favor reusing code vs adding new whenever possible.

nsarupr: Addressed this.

opta/core/azure.py
nsarupr: Removed Azure support for now.

opta/commands/config.py
nsarupr: Removed it.

nsarupr:
@RemyDeWolf I have tried best to address the comments left. 
Please let me know if there is anything which may have been overlooked.

opta/core/aws.py
```diff
@@ -167,6 +161,35 @@ def force_delete_terraform_lock_id(self) -> None:
...
+                    structured_config = cls._download_remote_config(
+                        s3, aws_bucket, data["Key"]
+                    )
+                    if structured_config:
+                        detailed_configs[data["Key"][len(prefix) :]] = structured_config
+                opta_config_detailed_map[aws_bucket] = detailed_configs
+        return opta_config_detailed_map
+
+    @staticmethod
+    def _get_bucket_list() -> List[str]:
```
RemyDeWolf, line 183: this method is also doing some filtering, so the name should capture this information `_list_opta_buckets`

opta/core/aws.py
```diff
@@ -167,6 +161,35 @@ def force_delete_terraform_lock_id(self) -> None:
             Key={"LockID": {"S": f"{bucket}/{self.layer.name}"}},
         )
 
+    @classmethod
+    def get_all_remote_configs(cls) -> Dict[str, Dict[str, "StructuredConfig"]]:
+        prefix = "opta_config/"
+        s3 = boto3.client("s3")
+        opta_config_detailed_map = {}
```
RemyDeWolf, line 168: let's use consistent naming all over, this method uses `StructuredConfig` and `remote_configs`
`detailed_map` `detailed_config` seem like similar concepts but using different names make the code harder to read.

opta/core/gcp.py
```diff
@@ -85,14 +86,7 @@ def get_remote_config(self) -> Optional["StructuredConfig"]:
...
         bucket_object = gcs_client.get_bucket(bucket)
-        try:
-            blob = storage.Blob(config_path, bucket_object)
-            return json.loads(blob.download_as_text())
-        except Exception:  # Backwards compatibility
-            logger.debug(
-                "Could not successfully download and parse any pre-existing config"
-            )
-            return None
+        return self._download_remote_config(bucket_object, config_path)
```
RemyDeWolf, line 89: 👍

opta/core/gcp.py
```diff
@@ -173,3 +167,32 @@ def bucket_exists(self, bucket_name: str) -> bool:
...
+            ):
+                structured_config = cls._download_remote_config(bucket, response.name)
+                if structured_config:
+                    detailed_config[response.name[len(prefix) :]] = structured_config
+            if detailed_config:
+                opta_config_detailed_map[bucket.name] = detailed_config
+        return opta_config_detailed_map
+
+    @staticmethod
+    def _download_remote_config(bucket: Bucket, key: str) -> Optional["StructuredConfig"]:
```
RemyDeWolf, line 190: key is a `path`

tests/core/test_aws.py
```diff
@@ -70,3 +73,63 @@ def test_force_delete_terraform_lock_id(
             ],
             Key={"LockID": {"S": f"{aws_layer.state_storage()}/{aws_layer.name}"}},
         )
+
+    def test_get_detailed_config_map_configuration_present(
```
RemyDeWolf, line 77: also rename the tests accordingly `test_new _name`
otherwise if the test fail someone will not know what is broken from the name

opta/core/cloud_client.py
```diff
@@ -28,3 +28,7 @@ def delete_remote_state(self) -> None:
     @abstractmethod
     def get_terraform_lock_id(self) -> str:
         raise NotImplementedError()
+
+    @classmethod
+    def get_all_remote_configs(cls) -> Dict[str, Dict[str, "StructuredConfig"]]:
```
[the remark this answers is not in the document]
nsarupr: Then we can not print the Bucket name in the `opta config view`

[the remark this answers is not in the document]
RemyDeWolf: that's fine it was not in the requirement.

opta/core/gcp.py
nsarupr: In bucket terminology, `key` is better understood as compared to the path.

opta/commands/config.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def __get_cloud_client(cloud: str) -> CloudClient:
+    cloud_client: CloudClient
+    if cloud.lower() == "aws":
+        cloud_client = AWS()
+    elif cloud.lower() == "google":
+        cloud_client = GCP()
+
+    return cloud_client
```
RemyDeWolf, line 54: JD needed something similar, we can consolidate such methods in the same place
https://github.com/run-x/opta/blob/3960cd526ce82299dfeab42d366d7b2da48cbd7d/opta/layer.py#L176

opta/core/aws.py
```diff
@@ -167,6 +162,33 @@ def force_delete_terraform_lock_id(self) -> None:
...
+                    structured_config = self._download_remote_config(
+                        s3, aws_bucket, data["Key"]
+                    )
+                    if structured_config:
+                        configs[data["Key"][len(prefix) :]] = structured_config
+                remote_configs[aws_bucket] = configs
+        return remote_configs
+
+    @staticmethod
+    def _get_opta_bucket() -> List[str]:
```
RemyDeWolf, line 183: if more than one object can be returned let's use plural `_get_opta_buckets`
for example see the naming for the official api below `s3.list_buckets()`

opta/core/azure.py
```diff
@@ -170,3 +171,6 @@ def bucket_exists(self, bucket_name: str) -> bool:
             return True
         except ResourceNotFoundError:
             return False
+
+    def get_all_remote_configs(self) -> Dict[str, Dict[str, "StructuredConfig"]]:
+        raise UserErrors("Feature Unsupported for Azure")
```
RemyDeWolf, line 176: instead of changing local and azure, it's easier to just raise the exception in the abstract method
like what is done here https://github.com/run-x/opta/blob/3960cd526ce82299dfeab42d366d7b2da48cbd7d/opta/core/cloud_client.py#L34

opta/core/cloud_client.py
```diff
@@ -1,13 +1,14 @@
...
 if TYPE_CHECKING:
     from opta.layer import Layer, StructuredConfig
 
 
 class CloudClient(ABC):
-    def __init__(self, layer: "Layer"):
-        self.layer = layer
+    def __init__(self, layer: Optional["Layer"] = None):
+        if layer:
+            self.layer = layer
```
RemyDeWolf, line 11: 👍

opta/core/gcp.py
RemyDeWolf: that's fine, it looks like aws uses key with gcp uses name.
gcp storage.Blob() has a param `encryption_key` so that's why I was suggesting a different name. but I didn't realize that AWS was using key.

RemyDeWolf (commented):
we are getting close, JD added a new method that we should reuse and some minor comments to address.

opta/core/azure.py
nsarupr: Raising a user error is much better, because in hierarchy it is handled with a Slack support message, otherwise we'll have to duplicate that part of code again.

opta/commands/config.py
```diff
@@ -0,0 +1,54 @@
...
+@config.command()
+@cloud_option
+def view(cloud: str) -> None:
+    """
+    View the opta configuration file for a given cloud provider
+
+    Use the current cloud credentials to fetch the remote opta config file
+    """
+    if cloud.lower() == "azurerm":
+        raise UserErrors("Currently AzureRM isn't supported.")
```
NitinAgg, line 36: nit: Currently Azure is not supported for this comment. As Opta's Azure support is still in Beta

opta/commands/config.py
NitinAgg: +1

opta/core/aws.py
```diff
@@ -30,8 +31,9 @@ class AwsArn(TypedDict):
 
 
 class AWS(CloudClient):
-    def __init__(self, layer: "Layer"):
-        self.region = layer.root().providers["aws"]["region"]
+    def __init__(self, layer: Optional["Layer"] = None):
+        if layer:
```
NitinAgg, line 35: nit `if layer is not None`

opta/core/azure.py
NitinAgg: We can define a new UserError (we have a couple already), but specific to no Azure support

opta/core/cloud_client.py
```diff
@@ -1,13 +1,14 @@
...
 
 if TYPE_CHECKING:
     from opta.layer import Layer, StructuredConfig
 
 
 class CloudClient(ABC):
-    def __init__(self, layer: "Layer"):
-        self.layer = layer
+    def __init__(self, layer: Optional["Layer"] = None):
+        if layer:
```
NitinAgg, line 10: nit: `if layer is not None`

NitinAgg (commented):
Minor comments.

opta/core/cloud_client.py
nsarupr: They are both the same things.

opta/core/azure.py
RemyDeWolf: @nsarupr the config command already raises a user error for azure https://github.com/run-x/opta/blob/feat/nilesh/runx-1121-opta-get-service/opta/commands/config.py#L35-L36

RemyDeWolf: this would work too raise UserError("This method is not supported by this cloud provider")

opta/commands/config.py
nsarupr: Sure.

nsarupr: For this one, we can have another PR and figure out where can we put this method.

NitinAgg (approved):

PR #727, opened 2022-02-22
Allowing people to specify cron tasks for k8s-service

juandiegopalomino:
# Description
Turned out to be pretty straightforward. Details are in the markdown.

# Safety checklist

manually

### DIFF
--- a/config/registry/schemas/common-types/cron.json
+++ b/config/registry/schemas/common-types/cron.json
@@ -1,0 +1,23 @@
+{
+  "$id": "https://app.runx.dev/common-types/cron",
+  "type": "object",
+  "properties": {
+    "commands": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "The command array to execute",
+      "pattern": "^[a-zA-Z0-9-]+$"
+    },
+    "schedule": {
+      "type": "string",
+      "description": "The crontab schedule for this job",
+      "pattern": "^[a-zA-Z0-9 /*-]+$"
+    },
+    "env_vars": {
+      "type": "object"
+    }
+  },
+  "required": ["commands", "schedule"]
+}
--- a/config/registry/schemas/common-types/cron.json
+++ b/config/registry/schemas/common-types/cron.json
@@ -7,13 +7,11 @@
       "items": {
         "type": "string"
       },
-      "description": "The command array to execute",
-      "pattern": "^[a-zA-Z0-9-]+$"
+      "description": "The command array to execute"
     },
     "schedule": {
       "type": "string",
-      "description": "The crontab schedule for this job",
-      "pattern": "^[a-zA-Z0-9 /*-]+$"
+      "description": "The crontab schedule for this job"
     },
     "env_vars": {
       "type": "object"
--- a/modules/aws_k8s_service/aws-k8s-service.md
+++ b/modules/aws_k8s_service/aws-k8s-service.md
@@ -122,9 +122,7 @@ Opta gives you the option of adding a list of cron jobs to run as part of this s
 field which a user can fill with entries for each con job in mind. Each entry must specify a command in array format
 (for most cases simply specify the shell you wish to use, the `-c` flag and the executable to run), as well as a
 schedule following the [Cron Syntax](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax).
-A user may additionally specify extra environment variables to use. All links and environment variables specified
-for the server will also be passed on to the cron tasks. The cron jobs will use the same resource requests/limits as the
-servers.
+The cron jobs will use the same resource requests/limits as the servers.
 
 For example, here is a service which has a cron job that runs every minute and simply outputs "Hello world!" to stdout:
 
@@ -141,8 +139,6 @@ For example, here is a service which has a cron job that runs every minute and s
         - -c
         - 'echo "Hello world!"'
         schedule: "* * * * *"
-        env_vars:
-          CRON: "1"
 ```
 
 {{% alert title="Pure Cron Jobs" color="info" %}}
--- a/modules/aws_k8s_service/aws-k8s-service.md
+++ b/modules/aws_k8s_service/aws-k8s-service.md
@@ -115,3 +115,32 @@ said nodes unless they have a matching toleration. Simply provide a list of desi
 
 Please refer to the taints specified in your environment Opta manifests to know what matching tolerations are right 
 for you.
+
+### Crons
+
+Opta gives you the option of adding a list of cron jobs to run as part of this service. This is done via the `crons`
+field which a user can fill with entries for each con job in mind. Each entry must specify a command in array format
+(for most cases simply specify the shell you wish to use, the `-c` flag and the executable to run), as well as a
+schedule following the [Cron Syntax](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax).
+A user may additionally specify extra environment variables to use. All links and environment variables specified
+for the server will also be passed on to the cron tasks. The cron jobs will use the same resource requests/limits as the
+servers.
+
+For example, here is a service which has a cron job that runs every minute and simply outputs "Hello world!" to stdout:
+
+```yaml
+  - name: app
+    type: k8s-service
+    image: AUTO
+    healthcheck_path: "/get"
+    port:
+      http: 80
+    crons:
+      - commands:
+        - /bin/sh
+        - -c
+        - 'echo "Hello world!"'
+        schedule: "* * * * *"
+        env_vars:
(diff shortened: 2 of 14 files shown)


modules/opta-k8s-service-helm/templates/crons.yaml
```diff
@@ -0,0 +1,62 @@
...
+          tolerations:
+          {{ range $toleration :=  $.Values.tolerations }}
+          - key: {{ $toleration.key | quote }}
+            value: {{ default "opta" $toleration.value | quote }}
+            effect: {{ default "NoSchedule" $toleration.effect | quote }}
+          {{ end }}
+          {{- end }}
+          containers:
+            - name: cron
+              image: "{{ $.Values.image }}"
```
bigbitbus, line 28: So if we have a "cron" in the opta service file in a k8s-service module, then we are installing a "deployment" (the current behavior) and this "cronjob" resource into Kubernetes with the identical image and environment in both the deployment and the cronjob. That should be useful in e.g. running migrations or backups etc. periodically. 

But...then we have wedded the two aspects; so when one changes, the other does too. Should be fine but its a little unconventional to combine them. Although I cant think of another way given that the k8s-service deploy command accepts a single image name/tag only so for example we dont have the possibility of creating two different k8s-service modules, one for deployment and one for cronjob.  

I'd rather use the k8s-resource and define a separate cronjob there, but thats just a choice.

modules/opta-k8s-service-helm/templates/crons.yaml
```diff
@@ -0,0 +1,62 @@
...
+              imagePullPolicy: Always
+              command:
+              {{ range $command := $cron.commands }}
+              - {{ $command | quote }}
+              {{ end }}
+              env:
+              - name: RDS_CA_PATH # This is the path to the public key for the docdb tls
```
bigbitbus, line 35: Why is this specific RDS_CA_PATH a special case to be defined inside the template; separate from the other envVars?

bigbitbus (commented):
Looks like a practical solution; some comments to think of included.

modules/opta-k8s-service-helm/templates/crons.yaml
juandiegopalomino: needed for documentdb from way back in the day. Super annoying and we might want to consider removing it lol

modules/opta-k8s-service-helm/templates/crons.yaml
juandiegopalomino: Great insight, 2 points:
1. So we can totally just have a pure cronjob service by specifying the min/max containers to be 0-- then only the cronjobs would run!
2. The special treatment we've given to the image has bitten us so much in the past, but it looks from today's discussion and plans for the onsite that we'll be trying to purge it very soon

modules/aws_k8s_service/aws-k8s-service.md
```diff
@@ -115,3 +115,32 @@ said nodes unless they have a matching toleration. Simply provide a list of desi
...
+    port:
+      http: 80
+    crons:
+      - commands:
+        - /bin/sh
+        - -c
+        - 'echo "Hello world!"'
+        schedule: "* * * * *"
+        env_vars:
+          CRON: "1"
```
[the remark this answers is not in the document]
juandiegopalomino: if for whatever reason they want to pass extra env vars.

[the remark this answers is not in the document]
bigbitbus: ok.

modules/opta-k8s-service-helm/templates/crons.yaml
bigbitbus: Ok

bigbitbus (approved):
approved after comments discussed.

config/registry/schemas/common-types/cron.json
```diff
@@ -0,0 +1,23 @@
...
+      },
+      "description": "The command array to execute",
+      "pattern": "^[a-zA-Z0-9-]+$"
+    },
+    "schedule": {
+      "type": "string",
+      "description": "The crontab schedule for this job",
+      "pattern": "^[a-zA-Z0-9 /*-]+$"
+    },
+    "env_vars": {
```
RemyDeWolf, line 18: I'm questioning if we need this?
why not reuse the same env_vars that is set on the main container?
would that make it so the user doesn't need to update the env vars at 2 places?
or do you have a good use case where we need 2 different set of env vars?

juandiegopalomino: I was thinking sometimes a user would like to specify that this is a cron job, but I'm happy to remove it if you think it's a bad idea

config/registry/schemas/common-types/cron.json
```diff
@@ -0,0 +1,23 @@
...
+      "items": {
+        "type": "string"
+      },
+      "description": "The command array to execute",
+      "pattern": "^[a-zA-Z0-9-]+$"
+    },
+    "schedule": {
+      "type": "string",
+      "description": "The crontab schedule for this job",
+      "pattern": "^[a-zA-Z0-9 /*-]+$"
```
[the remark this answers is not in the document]
juandiegopalomino: done

modules/aws_k8s_service/aws-k8s-service.md
```diff
@@ -115,3 +115,32 @@ said nodes unless they have a matching toleration. Simply provide a list of desi
 
 Please refer to the taints specified in your environment Opta manifests to know what matching tolerations are right 
 for you.
+
+### Crons
+
+Opta gives you the option of adding a list of cron jobs to run as part of this service. This is done via the `crons`
```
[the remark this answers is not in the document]
juandiegopalomino: done

config/registry/schemas/common-types/cron.json
```diff
@@ -0,0 +1,23 @@
...
+  "$id": "https://app.runx.dev/common-types/cron",
+  "type": "object",
+  "properties": {
+    "commands": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      },
+      "description": "The command array to execute",
+      "pattern": "^[a-zA-Z0-9-]+$"
```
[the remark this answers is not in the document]
juandiegopalomino: done

modules/opta-k8s-service-helm/templates/crons.yaml
juandiegopalomino: done

config/registry/schemas/common-types/cron.json
RemyDeWolf: my preference would be to remove it (and from doc) to keep it as simple as possible. if a user comes with this requirement it's easy to add 👍

modules/gcp_k8s_service/gcp-k8s-service.md
```diff
@@ -116,3 +116,36 @@ said nodes unless they have a matching toleration. Simply provide a list of desi
 Please refer to the taints specified in your environment Opta manifests to know what matching tolerations are right
 for you.
 
+### Crons
```
RemyDeWolf, line 119: rename this title too

RemyDeWolf (approved):
2 minor follow up comments but overall looks great

config/registry/schemas/common-types/cron.json
```diff
@@ -0,0 +1,21 @@
+{
+  "$id": "https://app.runx.dev/common-types/cron",
```
RemyDeWolf, line 2: rename the file and here too?

config/registry/schemas/common-types/cron.json
RemyDeWolf: did you mean to remove it?

modules/aws_k8s_service/aws-k8s-service.yaml
```diff
@@ -190,7 +190,15 @@ inputs:
     validator: list(include('toleration'), required=False)
     description: Taint tolerations to add to the pods.
     default: []
+  - name: cron_jobs
+    user_facing: true
+    validator: list(include('cron'), required=False)
+    description: A list of cronjobs to execute as part of this service
+    default: []
 extra_validators:
+  cron:
```
RemyDeWolf, line 199: rename here also?

modules/gcp_k8s_service/gcp-k8s-service.yaml
```diff
@@ -181,7 +181,15 @@ inputs:
     validator: list(include('toleration'), required=False)
     description: Taint tolerations to add to the pods.
     default: []
+  - name: cron_jobs
+    user_facing: true
+    validator: list(include('cron'), required=False)
+    description: A list of cronjobs to execute as part of this service
+    default: []
 extra_validators:
+  cron:
```
RemyDeWolf, line 190: rename here?

PR #728, opened 2022-02-22
Adding label to kube-system namespace

bigbitbus:
# Description
Add a label to kube-system because of this:
https://linkerd.io/2.10/features/ha/#proxy-injector-failure-policy

# Safety checklist

Circle Testing 
https://app.circleci.com/pipelines/github/run-x/opta?branch=feature%2Fadd-kube-system-label&filter=all

Also, verified that kube-system namespace has the required label in EKS.

### DIFF
--- a/modules/aws_k8s_base/aws_k8s_base.py
+++ b/modules/aws_k8s_base/aws_k8s_base.py
@@ -7,7 +7,12 @@ from ruamel.yaml.compat import StringIO
 
 from modules.base import AWSK8sModuleProcessor, K8sBaseModuleProcessor
 from opta.core.aws import AWS
-from opta.core.kubernetes import list_namespaces, load_opta_kube_config, set_kube_config
+from opta.core.kubernetes import (
+    add_linkerd_label_to_kubesystem,
+    list_namespaces,
+    load_opta_kube_config,
+    set_kube_config,
+)
 from opta.exceptions import UserErrors
 from opta.utils import yaml
 
@@ -74,7 +79,9 @@ class AwsK8sBaseProcessor(AWSK8sModuleProcessor, K8sBaseModuleProcessor):
 
     def pre_hook(self, module_idx: int) -> None:
         set_kube_config(self.layer)
+        add_linkerd_label_to_kubesystem()
         list_namespaces()
+
         super(AwsK8sBaseProcessor, self).pre_hook(module_idx)
 
     def post_hook(self, module_idx: int, exception: Optional[Exception]) -> None:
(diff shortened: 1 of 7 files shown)


RemyDeWolf (commented):
After reading the linkerd doc, my understanding is that it would only add the proxy if 
`linkerd.io/inject: enabled` is set. But we don't set this to the kube-system namespace right?

Also, the issue with this change is that it won't work with terraform generator, because terraform would not add the label. Would that be a blocker?
-> We can have the label done by helm, here is an example https://gist.github.com/kvudata/12ba57ae1e7f01799aaa7f36350a9b2e

bigbitbus:
> After reading the linkerd doc, my understanding is that it would only add the proxy if `linkerd.io/inject: enabled` is set. But we don't set this to the kube-system namespace right?
> 
> Also, the issue with this change is that it won't work with terraform generator, because terraform would not add the label. Would that be a blocker? -> We can have the label done by helm, here is an example https://gist.github.com/kvudata/12ba57ae1e7f01799aaa7f36350a9b2e

This label requirement came from a customer incident @rabbitfang addressed and then created the ticket (see description here: https://run-x.atlassian.net/browse/RUNX-1160); he suggests we add the label config.linkerd.io/admission-webhooks=disabled on the kube-system namespace and that seemed to have fixed the issue of the webhook.

True, the terraform will not contain this because we are directly invoking the API instead of using Terraform; I feel running a helm chart via terraform to just set one label is going to be much slower than invoking this API call, for every user/cluster. We haven't solved the question of what to do with all this Python glue code (within processors) in Terraform code generation; once we have a solution to that it should wrap up this issue. 

Also, the approach in the github gist will require helm to "adopt" or "import" the kube-system namespace (As helm does not work out-of-the-box with pre-existing resources); I am a little uncomfortable touching the kube-system namespace that is essentially managed by AKS/EKS/GKE any more than needed.

Lets chat more if you want.

modules/aws_k8s_base/aws_k8s_base.py
```diff
@@ -74,7 +79,9 @@ def process(self, module_idx: int) -> None:
 
     def pre_hook(self, module_idx: int) -> None:
         set_kube_config(self.layer)
+        add_linkerd_label_to_kubesystem()
         list_namespaces()
```
rabbitfang, line 83:
```suggestion
        list_namespaces()
        add_linkerd_label_to_kubesystem()
```

We are using `list_namespaces` to verify connectivity to the cluster, so we shouldn't try adding the label to the namespace until we are sure we can connect. Ditto for later instances.

rabbitfang (changes requested):

bigbitbus:
closing for #732 instead.

PR #729, opened 2022-02-24
Allowing min and max containers to be set to zero

juandiegopalomino:
# Description
For pure cron job k8s services.

# Safety checklist

manually


modules/opta-k8s-service-helm/templates/deployment.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if eq (len .Values.persistentStorage) 0 }}
 {{- if .Values.deployPods }}
+{{- if not (and (eq "0" .Values.autoscaling.minReplicas) (eq "0" .Values.autoscaling.maxReplicas)) }}
```
RemyDeWolf, line 3: in the helm chart we can probably ignore minReplicas to make the decision `not (eq "0" .Values.autoscaling.maxReplicas)` - easier to read

RemyDeWolf (approved):

PR #732, opened 2022-02-24
Added extra proxy injector exclusion as kube-system in values

bigbitbus:
# Description
Use matchlabel to keep kube-system out of the linkerd proxy. Alternative to https://github.com/run-x/opta/pull/728

# Safety checklist

Circle CI


RemyDeWolf (approved):
this is a great way to solve this issue 👏

PR #735, opened 2022-02-25
Terraform Generate for Opta modules used by AWS/GCP

nsarupr:
# Description
- S3
- DocDB
- Helm-Chart
- MySQL
- Redis
- PostGres
- AWS IAM Role
- AWS IAM User
- AWS SNS
- AWS SQS
- GCP Service Account
- GCP Storage

# Safety checklist

https://app.circleci.com/pipelines/github/run-x/opta/1623/workflows/b85c1790-129c-47e9-97c5-bb7178394a10
https://app.circleci.com/pipelines/github/run-x/opta/1655/workflows/39f763d3-a38f-4760-9800-c70c14e52961


.circleci/config.yml
```diff
@@ -424,25 +424,113 @@ jobs:
...
+            ./dist/opta/opta generate-terraform \
+            --config ./.circleci/ci-tests/terraform-generated-aws/module-docdb.yaml \
+            --directory $HOME/generated-tf-files-aws \
+            --auto-approve
+      - run:
+          name: "Plan Terraform Generated for DocDB"
+          command: |
+            cd $HOME/generated-tf-files-aws
+            terraform init
+            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.tgdocdb
```
RemyDeWolf, line 468: anyway we can avoid repeating this code?

this post is interesting, I wonder if we could something similar:
https://discuss.circleci.com/t/circleci-orb-with-array-parameters-and-for-loop/36964/2

```
          command: |
            printf "Validating terraform code\n"
            for i in <<parameters.modules>>; do
               # find  opta file in module directory.
               # if not found, generate a default opta file to use with the module to test
               # run opta generate-terraform
               # run terraform init, plan
            done
```

.circleci/ci-tests/terraform-generated-aws/module-docdb.yaml
```diff
@@ -0,0 +1,9 @@
+environments:
+  - name: env-tg
+    path: "environment.yaml"
+    variables:
+      max_containers: 2
+name: module-docdb
+modules:
+  - name: tgdocdb
+    type: aws-documentdb
```
RemyDeWolf, line 9: if the module doesn't have any mandatory attribute, then this file could be generated automatically by circleci.
this will also make adding new test case easier for such modules.

nsarupr: This is something which can be looked up on later, if and when modules are added more frequently.
As a basic test, I have only added the modules which are used much more frequently.

RemyDeWolf: ok, it can be added later, but we need to test all the modules with `tf plan` as part of this effort.
the goal is to find the issues and either mark these modules with error as unsupported for export and then switch the command from beta to available.

.circleci/config.yml
nsarupr: How about we do something like this?

```yaml
version: 2.1
commands: # a reusable command with parameters
  opta-module-with-terraform-generate:
    parameters:
      module:
        default: ""
        type: string
    steps:
      - run:
          name: Generate Terraform Language Files for << parameters.module >>
          command: |
            ./dist/opta/opta generate-terraform \
            --config ./.circleci/ci-tests/terraform-generated-aws/module-docdb.yaml \
            --directory $HOME/generated-tf-files-aws \
            --auto-approve
      - run:
          name: Terraform Plan for << parameter.module >>
          commamd: |
            cd $HOME/generated-tf-files-aws
            terraform init
            terraform plan -compact-warnings -lock=false -input=false -out=tf.plan -target=module.<< parameter.module >>
jobs:
  test-terraform-module:
    executor: ubuntu-machine
    steps:
      - opta-module-with-terraform-generate:
          module: "docdb"
      - opta-module-with-terraform-generate:
          module: "s3"
workflows:
  my-workflow:
    jobs:
      - test-terraform-module
```

RemyDeWolf: @nsarupr sorry for responding only now, it looks amazing, great case of reuse vs duplicate 💯

.circleci/config.yml
```diff
@@ -420,25 +445,55 @@ jobs:
...
+      - opta-module-with-terraform-generate:
+          module: "pg"
+      - opta-module-with-terraform-generate:
+          module: "awsiamrole"
+      - opta-module-with-terraform-generate:
+          module: "awsiamuser"
+      - opta-module-with-terraform-generate:
+          module: "sns"
+      - opta-module-with-terraform-generate:
+          module: "sqs"
```
RemyDeWolf, line 496: 👍

RemyDeWolf (commented):
looks good!

can you add a readme, to describe how to add a new generate terraform for a module, so anybody would be able to add such test with new modules.

what is the plan for the modules that are not in this PR?

nsarupr:
> looks good!
> 
> can you add a readme, to describe how to add a new generate terraform for a module, so anybody would be able to add such test with new modules.
> 
> what is the plan for the modules that are not in this PR?

As for the Azure modules, we are not adding anything as of now.
For the other modules, such as DNS and other dependents we can have a call on what needs to be done.

RemyDeWolf:
@NitinAgg I didn't look at every module, but I can see some modules not covered, ex: dynamodb, awsses.
what's the plan for such modules?

nsarupr:
> @nsarupr I didn't look at every module, but I can see some modules not covered, ex: dynamodb, awsses. what's the plan for such modules?

SES has a dependency on DNS module (https://docs.opta.dev/reference/aws/environment_modules/aws-ses/)
DynamoDB I'm in the process of adding.

RemyDeWolf:
Got it, then let's add this flag to the module that depends on dns, we can make this change with this PR
https://github.com/run-x/opta/blob/04cee831cfb29d5166f95bc3a51a939d4e4319cc/modules/runx/runx.yaml#L3

RemyDeWolf (approved):

.circleci/config.yml
```diff
@@ -451,6 +516,8 @@ jobs:
           command: |
             cd $HOME/generated-tf-files-aws
             terraform init
+            terraform refresh
+            terreaform state rm module.base.aws_ebs_default_kms_key.default
```
RemyDeWolf, line 520: let's add one comment about why we added this

RemyDeWolf (approved):
looks, just add one comment next to the workaround we added for destroy to work so we will remember why we did this

PR #736, opened 2022-02-25
Fix retool example code

bigbitbus:
# Description
Fixes bug, updates helm chart version for retool

# Safety checklist

manual

### DIFF
--- a/examples/retool/retool-aws.yaml
+++ b/examples/retool/retool-aws.yaml
@@ -20,7 +20,7 @@ modules:
     name: console
     chart: retool
     repository: https://charts.retool.com
-    chart_version: "4.7.0"
+    chart_version: "4.8.0"
     values:
       postgresql:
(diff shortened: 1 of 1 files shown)


nsarupr (approved):
Defaults Chart values for [Retool Helm](https://github.com/tryretool/retool-helm) for Ingress as mentioned [here](https://github.com/tryretool/retool-helm/blob/main/values.yaml) 
```yaml
ingress:
  enabled: false
  labels: {}
  annotations: {}
  # kubernetes.io/ingress.class: nginx
  # kubernetes.io/tls-acme: "true"
  # configures the hostname e.g. retool.example.com
  hostName:
  tls:
  # - secretName: retool.example.com
  #   hosts:
  #     - retool.example.com
  # servicePort: service-port
  pathType: ImplementationSpecific
```

The change looks good to me.

PR #738, opened 2022-02-25
Made cluster name passed in and added does_cluster_exist method

juandiegopalomino:
# Description
Also had to do some updates on AKS for this to work.
@RemyDeWolf @bigbitbus I confirmed it fixed the eks creation issue

# Safety checklist

Manually


modules/aws_eks/aws-eks.yaml
```diff
@@ -14,6 +14,10 @@ inputs:
     user_facing: false
     description: Opta Module name
     default: None
+  - name: cluster_name
```
bigbitbus, line 17: We are changing the cluster name in the terraform files from 
name     = "opta-${var.layer_name}"

to

 name     = var.cluster_name

Have you confirmed that for pre-existing clusters the name doesn't change from what its set to? To prevent cluster re-creation?

modules/azure_aks/tf_module/variables.tf
```diff
@@ -5,9 +5,9 @@ data "azurerm_subscription" "current" {}
 data "azurerm_client_config" "current" {}
 
 data "azurerm_subnet" "opta" {
-  name                 = "opta-${var.env_name}-subnet"
+  name                 = var.private_subnet_name
```
bigbitbus, line 8: Have we confirmed pre-existing subnet names are not impacted via a test on a pre-existing cluster?

opta/core/kubernetes.py
```diff
@@ -387,12 +394,102 @@ def _aws_set_kube_config(layer: "Layer") -> None:
...
+        return _azure_does_cluster_exist(layer)
+    elif layer.cloud == "local":
+        return _local_does_cluster_exist(layer)
+    else:
+        raise Exception(
+            f"Unknown cloud {layer.cloud}. Can not handle determining if cluster exists"
+        )
+
+
+def _local_does_cluster_exist(layer: "Layer") -> bool:
```
bigbitbus, line 423: Consider caching since these functions are expensive API calls
https://docs.python.org/3/library/functools.html#functools.cache

bigbitbus (commented):
Left some comments, this is a good improvement

modules/aws_eks/aws-eks.yaml
juandiegopalomino: yes, I am absolutely certain

modules/azure_aks/tf_module/variables.tf
juandiegopalomino: yup! I tried it out with my pre-existing azure test cluster

opta/core/kubernetes.py
juandiegopalomino: I thought about it, but am not certain it is wise because our current troubles started because of similar caching

bigbitbus: ok, if the time hit is not worth the risk then lets pass.

bigbitbus (approved):

modules/aws_eks/aws-eks.yaml
```diff
@@ -14,6 +14,10 @@ inputs:
...
+    user_facing: false
```
RemyDeWolf, line 18: 👍

opta/core/kubernetes.py
```diff
@@ -387,12 +394,102 @@ def _aws_set_kube_config(layer: "Layer") -> None:
...
+    elif layer.cloud == "google":
+        return _gcp_does_cluster_exist(layer)
+    elif layer.cloud == "azurerm":
+        return _azure_does_cluster_exist(layer)
+    elif layer.cloud == "local":
+        return _local_does_cluster_exist(layer)
+    else:
+        raise Exception(
+            f"Unknown cloud {layer.cloud}. Can not handle determining if cluster exists"
+        )
```
RemyDeWolf, line 420: it's interesting, I didn't think we would change this part- we could still rely on terraform output to know if the cluster exist.
the issue we had was that it was bundled into `get_cluster_name` and would return None if terraform output didn't exist.

Did you find a corner case where we need to know if the cluster exist but where we won't have the terraform output?

modules/aws_eks/aws_eks.py
```diff
@@ -35,8 +39,7 @@ def post_delete(self, module_idx: int) -> None:
         purge_opta_kube_config(layer=self.layer)
 
     def pre_hook(self, module_idx: int) -> None:
-        cluster_name = get_cluster_name(self.layer.root())
-        if cluster_name is not None:
+        if does_cluster_exist(self.layer.root()):
```
RemyDeWolf, line 42: what if the cluster is gone but the cloudwatch log group was not deleted, are we still gonna delete them prior creating a new cluster?

opta/core/kubernetes.py
```diff
@@ -387,12 +394,102 @@ def _aws_set_kube_config(layer: "Layer") -> None:
...
+    return f"opta-{layer.root().name}"
+
+
+def does_cluster_exist(layer: "Layer") -> bool:
+    if layer.is_stateless_mode() is True:
+        if logger.isEnabledFor(DEBUG):
+            logger.debug(
+                "does_cluster_exist called in stateless mode, verify implementation. See stack trace below:"
+            )
+            traceback.print_stack()
```
RemyDeWolf, line 407: we can probably assume that cluster doesn't exist when running in stateless mode? False

opta/core/kubernetes.py
```diff
@@ -387,12 +394,102 @@ def _aws_set_kube_config(layer: "Layer") -> None:
...
-    outputs = get_terraform_outputs(layer)
-    cluster_name = outputs.get("parent.k8s_cluster_name") or outputs.get(
-        "k8s_cluster_name"
-    )
-    return cluster_name
+def get_cluster_name(layer: "Layer") -> str:
+    return f"opta-{layer.root().name}"
+
+
+def does_cluster_exist(layer: "Layer") -> bool:
```
RemyDeWolf, line 401: suggestion rename `cluster_exist`

opta/core/kubernetes.py
```diff
@@ -387,12 +394,102 @@ def _aws_set_kube_config(layer: "Layer") -> None:
...
+        return _aws_does_cluster_exist(layer)
+    elif layer.cloud == "google":
+        return _gcp_does_cluster_exist(layer)
+    elif layer.cloud == "azurerm":
+        return _azure_does_cluster_exist(layer)
+    elif layer.cloud == "local":
+        return _local_does_cluster_exist(layer)
+    else:
+        raise Exception(
+            f"Unknown cloud {layer.cloud}. Can not handle determining if cluster exists"
```
RemyDeWolf, line 419: to avoid this pattern, we can add an abstract method to CloudClient
https://github.com/run-x/opta/blob/main/opta/core/cloud_client.py#L8

it makes it clear about what needs to be implemented for each cloud.

but so also my other comment below, if we need this new cloud method or not.

opta/core/kubernetes.py
juandiegopalomino: That's actually why we had the issue in the first place-- we cached terraform state from before the k8s cluster creation, and so when we checked the output we thought that the cluster was not created. I thought we also agreed to make this does_cluster_exist function and that it must not rely on the terraform output?

opta/core/kubernetes.py
juandiegopalomino: I don't understand-- should we just return false?

modules/aws_eks/aws_eks.py
juandiegopalomino: er... yes-- this is what these lines of code are doing, right?

opta/core/kubernetes.py
juandiegopalomino: done!

modules/aws_eks/aws_eks.py
RemyDeWolf: got it

opta/core/kubernetes.py
RemyDeWolf: yes return false.

opta/core/kubernetes.py
RemyDeWolf: yes, I remember now - we have this issue where the terraform output is saved in memory, so calling the cloud api is good 👍 thanks for the refresher

opta/core/kubernetes.py
```diff
@@ -206,207 +182,37 @@ def load_opta_kube_config_to_default(layer: "Layer") -> None:
...
-    if cluster_name is None:
+    if layer.cloud == "aws":
+        return AWS(layer).cluster_exist()
+    elif layer.cloud == "google":
+        return GCP(layer).cluster_exist()
+    elif layer.cloud == "azurerm":
+        return Azure(layer).cluster_exist()
+    elif layer.cloud == "local":
+        return Local(layer).cluster_exist()
+    else:
```
RemyDeWolf, line 206: we should be able to do this
```
@classmethod
getCloudClient(cloud str) -> getCloudClient:
# to impl
```

```
getCloudClient(layer.cloud).cluster_exist()
```
and also

```
getCloudClient(layer.cloud).set_kube_config()
```

opta/core/kubernetes.py
juandiegopalomino: @RemyDeWolf took a lot of refactoring, but I think it's for the best-- take another look!

opta/core/kubernetes.py
juandiegopalomino: I'm a bit lost-- is there currently a get_cloud_client method/function somewhere?

RemyDeWolf: no, we can create it in `CloudClient` so we don't need to repeat this pattern

```
if layer.cloud == "aws":
        return AWS(layer).bla()
    elif layer.cloud == "google":
        return GCP(layer).bla()
    elif layer.cloud == "azurerm":
        return Azure(layer).bla()
    elif layer.cloud == "local":
        return Local(layer).bla()
    else:
   # error
````

would be `CloudClient.get_cloud_client(cloud).bla()

BTW: Nilesh needs one too for his PR
https://github.com/run-x/opta/blob/d34d89dba7acd4d242c49b7294da416bc6972647/opta/commands/config.py#L47

Comment (juandiegopalomino) [reply opta/core/kubernetes.py:206]: no, that would be a circular dependency-- it ought to be part of layer. Let me take care of it now

Comment (RemyDeWolf) [reply opta/core/kubernetes.py:206]: https://www.geeksforgeeks.org/factory-method-python-design-patterns/

File: opta/core/kubernetes.py [hunk 11]
```diff
@@ -37,10 +37,6 @@
 
 import opta.constants as constants
 from opta.constants import GENERATED_KUBE_CONFIG_DIR, REDS, yaml
-from opta.core.aws import AWS
-from opta.core.azure import Azure
-from opta.core.gcp import GCP
-from opta.core.local import Local
```
RemyDeWolf: it's nice that k8s is no longer aware of each cloud impl

RemyDeWolf (approved):

opta/core/kubernetes.py
juandiegopalomino: yup!

PR #745, opened 2022-03-02
Adding the ability to pass runtime variables to opta

juandiegopalomino:
# Description
As per the discussion in our internal notion doc. Only string values supported at the moment, but I confirmed that it should not be an issue with our flyte users

# Safety checklist

Ran manually like this
```yaml
name: live-example-dev
org_name: runx
providers:
  aws:
    region: us-east-1
    account_id: REDACTED
input_variables:
  - name: max_nodes
  - name: min_nodes
    default: 1
modules:
  - type: base
  - type: dns
    domain: baloney.runx.dev
    delegated: true
  - name: deployeruser
    type: aws-iam-user
    extra_iam_policies:
      - "arn:aws:iam::aws:policy/AdministratorAccess"
  - name: deployerrole
    type: aws-iam-role
    extra_iam_policies:
      - "arn:aws:iam::aws:policy/AdministratorAccess"
    allowed_iams:
      - "${{module.deployeruser.user_arn}}"
  - type: k8s-cluster
    k8s_version: "1.21"
    spot_instances: true
    max_nodes: 13
    enable_metrics: true
    node_launch_template:
      user_data: |
        MIME-Version: 1.0
        Content-Type: multipart/mixed; boundary="==MYBOUNDARY=="

        --==MYBOUNDARY==
        Content-Type: text/x-shellscript; charset="us-ascii"

        #!/bin/bash
        if ( lsblk | fgrep -q nvme1n1 ); then
           mkfs.ext4 /dev/nvme1n1
           systemctl stop docker
           mkdir -p /var/lib/docker
           mount /dev/nvme1n1 /var/lib/docker
           chmod 711 /var/lib/docker
           systemctl start docker
        fi
        --==MYBOUNDARY==--
  - type: k8s-base
    admin_arns:
      - "${{module.deployeruser.user_arn}}"
      - "${{module.deployerrole.role_arn}}"
    nginx_extra_tcp_ports:
      - 25565
      - 25575
  - type: helm-chart # <-- Add this for LogDNA support
    chart: agent
    repository: https://assets.logdna.com/charts
    chart_version: 203.1.0
    values:
      logdna:
        key: "871a9e2660a89a394a0caa456fbfb6a8"
        tags: aws-live-example
  - type: aws-nodegroup
    name: nodegroup1
    spot_instances: true
    min_nodes: "{vars.min_nodes}"
    max_nodes: "{vars.max_nodes}"
    taints:
      - key: blah1
      - key: blah2
        value: baloney
      - key: blah3
        value: baloney
        effect: "NO_EXECUTE"
```


bigbitbus (commented):
Looks good, good idea in centralizing the click options in the utils dir. Just a comment on the initial parsing since may folks make mistakes on the cli.

opta/input_variable.py
```diff
@@ -0,0 +1,76 @@
...
+                raise UserErrors(
+                    f"Input variable {input_variable.name} was expected, but not given"
+                )
+        return output
+
+    @staticmethod
+    def parse_string_to_dict(inputs: List[str]) -> Dict[str, str]:
+        output = {}
+        for input in inputs:
+            input_parts = input.split("=", 1)
```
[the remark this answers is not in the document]
juandiegopalomino: I don't think click will allow whitespaces to begin with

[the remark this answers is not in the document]
juandiegopalomino:
```bash
$ python ../opta/cli.py output -c aws-live-example.yaml --var max_nodes = 5
Usage: cli.py output [OPTIONS]
Try 'cli.py output -h' for help.

Error: Got unexpected extra arguments (= 5)
```

opta/commands/apply.py
```diff
@@ -88,6 +87,7 @@ def apply(
     test: bool,
     auto_approve: bool,
     detailed_plan: bool,
+    var: List[str],
```
RemyDeWolf, line 90: can you add an example below in `Examples:`?
may be an example with 2 vars to explain the expected syntax

opta/commands/apply.py
```diff
@@ -78,6 +74,9 @@
     default=False,
     help="Show full terraform plan in detail, not the opta provided summary",
 )
+@config_option
+@env_option
+@input_variable_option
```
RemyDeWolf, line 79: I wonder if we should also support using a env file to set these variables, similar to `-var-file` in terraform.

we already have a similar option when we set secrets
https://github.com/run-x/opta/blob/main/opta/commands/secret.py#L25
https://github.com/run-x/opta/blob/main/opta%2Fcore%2Fsecrets.py#L39

you could make an extra option  `@input_variable_file_option` and have an extra arg to `InputVariable.parse_string_to_dict(var, var_file)`

definitely not a must have for now but if you think this is useful, you can take a closer look

opta/commands/generate_terraform.py
```diff
@@ -98,9 +98,12 @@ def generate_terraform(
     config = check_opta_file_exists(config)
 
     pre_check()
+    input_variables = InputVariable.parse_string_to_dict(var)
```
RemyDeWolf, line 101: can you check if we can use this [callback](https://click.palletsprojects.com/en/8.0.x/options/#callbacks-for-validation) instead of having to do this call in every command?

```python
input_variable_option = click.option(
    "--var",
    "-v",
    multiple=True,
    default=[],
    action=;
    help="Add input variables to your yaml at run time (e.g. --var variable1=value1)",
    callback=parse_string_to_dict #would be called automatically, would need to be changed to work with api
)
```

juandiegopalomino: done

opta/commands/apply.py
juandiegopalomino: for later lol

opta/utils/clickoptions.py
```diff
@@ -8,6 +28,22 @@
...
+
+config_option = click.option(
+    "-c", "--config", default="opta.yaml", help="Opta config file", show_default=True
+)
+input_variable_option = click.option(
+    "--var",
+    "-v",
+    multiple=True,
+    default=[],
+    callback=parse_variables,
```
RemyDeWolf, line 43: 👍

RemyDeWolf (approved):
looks great!

PR #749, opened 2022-03-03
Fail nice on simultaneous calls from same directory

nsarupr:
# Description
Fail nicely when a user invokes simultaneous runs on the same directory.

# Safety checklist

Tested manually.


opta/opta_run_check.py
```diff
@@ -0,0 +1,41 @@
...
+        pass
+
+
+def opta_run_start() -> None:
+    if check_opta_running_file_exists():
+        raise UserErrors("Opta already running in the current directory.")
+    create_opta_running_file()
+
+
+def opta_run_end() -> None:
```
bigbitbus, line 40: Can we add a debugging step in the opta docs - sometimes when people do Ctrl+C, this file may be left in the .opta directory, and so its like a stale lock; people should know that they need to remove it for next run.

opta/opta_run_check.py
```diff
@@ -0,0 +1,41 @@
...
+def check_opta_running_file_exists() -> bool:
+    return glob.glob(OPTA_RUNNING_LOC_PREFIX + "*").__len__() > 0
+
+
+def create_opta_running_file() -> None:
+    with open(f"{OPTA_RUNNING_LOC_PREFIX}_{os.getpid()}", "w") as f:
+        f.write("")
+
+
+def remove_opta_running_file() -> None:
```
bigbitbus, line 27: We have some nice "handling" when user presses Ctrl+C; can we check if this file can be deleted when user presses Ctrl+C before opta exits?

bigbitbus (commented):
Thanks, some comments around handling the ctrl+c case and the opta error case.

opta/opta_run_check.py
nsarupr: Let me test the `Ctrl + C` case. But it should be caught under any circumstances, and since the `opta_run_end` is running in finally block, every graceful exit will run this LOC. 
Although the question is do we need to handle ungraceful exits, such as killing the Process directly or something...
Or have some sort of release mechanism?

opta/opta_run_check.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+import glob
+import os
+
+from opta.constants import HOME
+from opta.exceptions import UserErrors
+
+OPTA_RUNNING_LOC_PREFIX = os.path.join(
+    HOME, ".opta", f".opta_running_{os.getcwd().replace(os.path.sep, '_')}_"
```
RemyDeWolf, line 14: why not use the current working directory when you run opta?
this is usually how these files are located, ex: https://www.terraform.io/language/files/dependency-lock#lock-file-location

for us it could be `$CWD/.opta.lock`

opta/opta_run_check.py
```diff
@@ -0,0 +1,41 @@
...
+)
+
+
+def check_opta_running_file_exists() -> bool:
+    return glob.glob(OPTA_RUNNING_LOC_PREFIX + "*").__len__() > 0
+
+
+def create_opta_running_file() -> None:
+    with open(f"{OPTA_RUNNING_LOC_PREFIX}_{os.getpid()}", "w") as f:
```
RemyDeWolf, line 23: why do we need the pid here?
if we call it `.opta.lock` would that make the code and the user troubleshooting (manually delete) easier?

opta/opta_run_check.py
```diff
@@ -0,0 +1,41 @@
...
+def remove_opta_running_file() -> None:
+    try:
+        os.remove(f"{OPTA_RUNNING_LOC_PREFIX}_{os.getpid()}")
+    except FileNotFoundError:
+        pass
+
+
+def opta_run_start() -> None:
+    if check_opta_running_file_exists():
+        raise UserErrors("Opta already running in the current directory.")
```
RemyDeWolf, line 36: I haven't checked but only apply/destroy should be included in this check.
we can have a user running `opta apply` and on another terminal doing an `opta logs` to troubleshoot what's going on live, right?

opta/opta_run_check.py
nsarupr: Not using the `CWD`, just to keep things clean in user's directory.
No other reason for it.

opta/opta_run_check.py
nsarupr: The reason for this was that the graceful exits are a part of `cli` and not `apply` and destroy`.
So, if there is by any chance an ungraceful exit, the user would be stuck in a state, making user's job difficult for using opta.

opta/opta_run_check.py
nsarupr: Verified it. Works.

opta/opta_run_check.py
nsarupr: I didn't understand this thing properly. Do you mean, we should update this in the Documentation, or add it to the debug logs. Sorry about the confusion.

opta/opta_run_check.py
nsarupr: Since currently the `opta_release_lock` is in the `finally` block, if opta is run twice, it would not know whether it is releasing the lock on the current failed run or an already running one.

opta/opta_run_check.py
nsarupr: Check if this fits the use case: https://github.com/run-x/opta/pull/749/commits/6536250dcde5cab5d713657272a3798806d8a88c

opta/opta_lock.py
```diff
@@ -0,0 +1,44 @@
...
+    try:
+        os.remove(f"{OPTA_LOCK_LOC_PREFIX}_{os.getpid()}")
+        logger.debug(f"Removed Opta lock on Directory: {os.getcwd()}")
+    except FileNotFoundError:
+        return
+
+
+def opta_acquire_lock() -> None:
+    if check_opta_running_file_exists():
+        raise UserErrors("Opta already running in the current directory.")
```
RemyDeWolf, line 39: from this error message, let's give a workaround for the user on how to delete the lock file.
so they can fix themselves, if they see this image and for some reason opta is not running.
"Opta lock file found {lock_file }. If Opta is not already running, delete this file and try again"

RemyDeWolf (approved):
one more comment about the error message, but otherwise looks good

opta/commands/deploy.py
```diff
@@ -72,72 +73,78 @@ def deploy(
...
 
-    pre_check()
-
-    config = check_opta_file_exists(config)
-    if local:
-        config = _local_setup(config, image_tag=tag, refresh_local_env=True)
-    if not is_service_config(config):
-        raise UserErrors(
-            fmt_msg(
+    try:
```
juandiegopalomino, line 76: where is the exception block?

juandiegopalomino: I see, nvmnd

juandiegopalomino (approved):
Is there a way to force unlock?

nsarupr:
> Is there a way to force unlock?

If you are talking about the terraform lock which we force unlock, that's already there.
For this opta lock, we don't have a force unlock, but the procedure is provided to the user in Error message if this comes up.

https://github.com/run-x/opta/blob/bd7af5d3fce7b51d703d8b21a77970d8250f1da6/opta/opta_lock.py#L36

PR #750, opened 2022-03-03
Update regula version to 2.5.0

nsarupr:
# Description
Update the regula version.

# Safety checklist

Tested Regula manually on the System.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,8 +74,8 @@ jobs:
       - name: Install regula
         run: |
           cd modules
-          wget -O regula_1.0.0_Linux_x86_64.tar.gz https://github.com/fugue/regula/releases/download/v1.0.0/regula_1.0.0_Linux_x86_64.tar.gz
-          tar -xvf regula_1.0.0_Linux_x86_64.tar.gz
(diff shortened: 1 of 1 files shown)


bigbitbus (commented):
The new version causes some tests to fail; you may want to (1) either fix the RC on the failed tests - but this means testing that the change in TF is backward compatible and doesnt cause infra to be torn down and spun up OR (2) Add more exceptions to the config/config.rego.

nsarupr:
> The new version causes some tests to fail; you may want to (1) either fix the RC on the failed tests - but this means testing that the change in TF is backward compatible and doesnt cause infra to be torn down and spun up OR (2) Add more exceptions to the config/config.rego.

@bigbitbus for this, I am not sure which one would be the better way to go forward with, it depends upon the errors and how we want to get through these. Shall we have a discussion this monday, to figure out which ones do we need to waive and which ones do we need to add and make those backward compatible.

bigbitbus (approved):
lgtm

bigbitbus:
@nsarupr you may close this since regula has been removed from the repo and github.

rabbitfang:
Superseded by #837

PR #751, opened 2022-03-04
Add support for remote state for module API

rabbitfang:
# Description
Add support for remote state in the Module API apply.

# Safety checklist

Still needs some testing


opta/cloud/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+    if not cloud_id:
+        raise ValueError("No cloud provider configured")
+
+    if cloud_id == "aws":
+        config = layer.providers.aws
+        assert config is not None
+
+        return AWSProvider(config)
+
+    raise ValueError(f"Unknown cloud provider {cloud_id}")
```
RemyDeWolf, line 25: would that make sense to move this one to cloud_client?
https://github.com/run-x/opta/blob/main/opta%2Fcore%2Fcloud_client.py
since it looks like this is something that needs to be implemented for each cloud

also we recently added https://github.com/run-x/opta/blob/3960cd526ce82299dfeab42d366d7b2da48cbd7d/opta/layer.py#L176-L188

opta/cloud/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if cloud_id == "aws":
+        aws_config = layer.providers.aws
+        assert aws_config is not None
+
+        config.region = aws_config.region
+
+        return S3Store(config)
+
+    raise ValueError(f"Unknown cloud provider {cloud_id}")
```
RemyDeWolf, line 54: similar comment, everywhere we have if cloud is ... could be an abstract method to implement in `CloudClient`.
this will prevent having different cloud implementation in the same file.

opta/cloud/aws/state.py
```diff
@@ -0,0 +1,227 @@
...
+                    "When trying to determine the status of the state bucket, we got an "
+                    f"{error_code} error with the message {error_message}"
+                )
+
+        logger.debug("S3 bucket for terraform state not found, creating a new one")
+
+        create_kwargs: Dict[str, Any] = {}
+        if region != "us-east-1":
+            location = cast(BucketLocationConstraintType, region)
+            create_kwargs["CreateBucketConfiguration"] = {"LocationConstraint": location}
```
RemyDeWolf, line 120: why the special logic for not us-east-1, can we avoid this?

opta/cloud/aws/state.py
```diff
@@ -0,0 +1,227 @@
...
+
+        s3
+        s3.create_bucket(Bucket=bucket_name, **create_kwargs)
+        time.sleep(
+            10
+        )  # TODO: Change this for a "wait until bucket resource is available"
```
RemyDeWolf, line 126: possibly this https://stackoverflow.com/a/49817544
ok with sleep for now if you think that's enough time

opta/cloud/aws/state.py
rabbitfang: It's a limitation to the S3 API. Adding a comment about it

opta/cloud/aws/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+from .provider import AWSProvider
+from .state import S3Store
+
+__all__ = ["AWSProvider", "AWSProviderConfig", "S3Store"]
+
+
+def register(registry: Registry) -> None:
+    registry.register_provider("aws", AWSProvider, AWSProviderConfig)
+    registry.register_state_store("s3", _make_store, config=AWSProviderConfig)
+    registry.register_default_store("aws", "s3")
```
RemyDeWolf, line 16: since we will always need all these things we can have one method with mandatory attributes.
`registry.register("aws", AWSProvider, AWSProviderConfig, [store])`
default store could be the first element of the list

opta/cloud/aws/config.py
```diff
@@ -0,0 +1,32 @@
...
+        self.account_ids: Optional[List[str]] = None
+
+    @classmethod
+    def from_dict(cls, raw: dict) -> AWSProviderConfig:
+        try:
+            region = raw["region"]
+        except KeyError:
+            raise UserErrors("AWS region must be provided when using the AWS provider")
+
+        p = cls(region)
```
RemyDeWolf, line 24: let's avoid using single letter variable - this is python not go :)

opta/cloud/aws/state.py
```diff
@@ -0,0 +1,230 @@
...
+
+        contents = f.getvalue()
+
+        return contents
+
+    @property
+    def _bucket(self) -> str:
+        c = self.config
+
+        return f"opta-tf-state-{c.org_name}-{c.layer_name}"
```
RemyDeWolf, line 78: there was a recent change to add a suffix to the bucket name.
https://github.com/run-x/opta/blob/main/opta/layer.py#L603

the complexity is to support old and new names, check with @bigbitbus on this topic as well

opta/cloud/aws/state.py
```diff
@@ -0,0 +1,230 @@
...
+
+        try:
+            s3.get_bucket_encryption(Bucket=bucket_name)
+            return
+        except ClientError as e:
+            error_code = e.response["Error"]["Code"]
+            if error_code == "AuthFailure":
+                raise UserErrors(
+                    "The AWS Credentials are not configured properly.\n"
+                    "Visit https://docs.aws.amazon.com/sdk-for-java/v1/developer-guide/setup-credentials.html "
```
RemyDeWolf, line 94: why is this link related to java sdk?
is there a more generic link for this?

opta/cloud/aws/state.py
```diff
@@ -0,0 +1,230 @@
...
+                    "for more information."
+                )
+
+            if error_code == "AccessDenied":
+                raise UserErrors(
+                    f"We were unable to access the S3 bucket, {bucket_name} on your AWS account (opta needs this to store state).\n"
+                    "Possible Issues: \n"
+                    " - Bucket name is not unique and might be present in some other Account. Try updating the name in Configuration file to something else.\n"
+                    " - It could also mean that your AWS account has insufficient permissions.\n"
+                    "Please fix these issues and try again!"
```
RemyDeWolf, line 104: 👍  great error message

RemyDeWolf (approved):

opta/cloud/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+
+def cloud_provider_for_layer(layer: Layer) -> CloudProvider:
+    """
+    Returns a new CloudProvider based on the layer configuration.
+    """
+
+    cloud_id, config = layer.providers.cloud_id_and_config
+
+    if not cloud_id:
+        raise ValueError("No cloud provider configured")
```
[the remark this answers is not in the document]
rabbitfang: I used ValueError here since it provides more clarity on what the issue is, with the intention of converting to UserErrors up the stack.

opta/cloud/aws/__init__.py
rabbitfang: State storage is a separate concept from which cloud in which terraform is managing resources, so I kept them separate here. It also allows the implementations to be written separately (as I did for `local`, which only has a state storage implementation right now).

PR #754, opened 2022-03-07
Use K8s version from variables not data

nsarupr:
# Description
Data is available only during the apply phase, so we are getting a mismatch in the plan created and the plan executed. 
Using k8s version from `aws-eks` module output.

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/aws_eks/aws-eks.yaml
+++ b/modules/aws_eks/aws-eks.yaml
@@ -110,6 +110,9 @@ outputs:
   - name: k8s_node_group_security_id
     export: true
     description: The id of the security group of the cluster nodepools.
+  - name: k8s_version
+    export: true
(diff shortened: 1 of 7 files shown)


bigbitbus (approved):
lgtm, verified in the CI run 

https://app.circleci.com/pipelines/github/run-x/opta/1672/workflows/b0a6e21b-bed1-449c-bc3e-f9b9007ae399/jobs/4271?invite=true#step-110-257

bigbitbus (approved):
Confirmed, also in CI 
https://app.circleci.com/pipelines/github/run-x/opta/1672/workflows/b0a6e21b-bed1-449c-bc3e-f9b9007ae399/jobs/4271?invite=true#step-110-257

PR #755, opened 2022-03-07
Use K8s version from variables not data (#754)

bigbitbus:
* Use K8s version from variables not data

* fix lint and test case

# Description
Hotfix in v0.27.2

# Safety checklist
* [ x] This change is backwards compatible and safe to apply by existing users
* [ x] This change will NOT lead to data loss
* [ x] This change will NOT lead to downtime who already has an env/service setup

CI, manual test


nsarupr (approved):
Released as Version 0.27.2, this change has been verified. On ci and manually.

juandiegopalomino (approved):

PR #757, opened 2022-03-08
Adding timeouts for k8s services

juandiegopalomino:
# Description
Now folks can pass in timeouts to specify if deployments take longer than 5 minutes

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/aws_k8s_service/tf_module/variables.tf
+++ b/modules/aws_k8s_service/tf_module/variables.tf
@@ -207,3 +207,7 @@ variable "pod_annotations" {
   default     = {}
   description = "values to add to the pod annotations for the k8s-service pods"
 }
+variable "timeout" {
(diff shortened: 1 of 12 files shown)


modules/aws_k8s_service/tf_module/variables.tf
```diff
@@ -200,3 +200,8 @@ variable "tolerations" {
 variable "cron_jobs" {
   default = []
 }
+
+variable "timeout" {
+  type    = number
+  default = 300
```
RemyDeWolf, line 206: default to 600 here too?

NitinAgg: Or rather, not have the default here. Make this a required field in terraform

NitinAgg (approved):

PR #765, opened 2022-03-10
Adding ability to import/reference external tf states

juandiegopalomino:
# Description
See the docs

# Safety checklist

Manually

### DIFF
--- a/modules/external_state/external-state.json
+++ b/modules/external_state/external-state.json
@@ -1,7 +1,7 @@
 {
   "$id": "https://app.runx.dev/modules/external-state",
   "type": "object",
-  "description": "Plugs a custom helm chart into your Opta k8s cluster",
+  "description": "Adds the ability to refer to external, preexisting terraform states in opta.",
   "properties": {
     "backend_type": {
       "type": "string",
--- a/modules/external_state/external-state.json
+++ b/modules/external_state/external-state.json
@@ -1,0 +1,41 @@
+{
+  "$id": "https://app.runx.dev/modules/external-state",
+  "type": "object",
+  "description": "Plugs a custom helm chart into your Opta k8s cluster",
+  "properties": {
+    "backend_type": {
+      "type": "string",
+      "description": "The type of backend this external state is using"
+    },
+    "config": {
+      "type": "string",
+      "description": "The configuration to connect to this remote state",
+      "default": {}
+    },
+    "type": {
+      "description": "The name of this module",
+      "enum": [
+        "external-state"
+      ],
+      "default": "external-state",
+      "type": "string"
+    }
+  },
+  "required": [
+    "backend_type",
+    "config",
+    "type"
+  ],
+  "opta_metadata": {
+    "module_type": "service",
+    "clouds": [
+      "aws",
+      "azure",
+      "gcp",
+      "local"
+    ],
+    "name": "external-state",
+    "display_name": "external-state",
+    "tags": []
+  }
+}
--- a/modules/external_state/external-state.json
+++ b/modules/external_state/external-state.json
@@ -5,10 +5,10 @@
   "properties": {
     "backend_type": {
       "type": "string",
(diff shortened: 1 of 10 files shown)


modules/external_state/external-state.json
```diff
@@ -0,0 +1,41 @@
...
+  "$id": "https://app.runx.dev/modules/external-state",
+  "type": "object",
+  "description": "Adds the ability to refer to external, preexisting terraform states in opta.",
+  "properties": {
+    "backend_type": {
+      "type": "string",
+      "description": "The type of backend this external state is using"
+    },
+    "config": {
+      "type": "string",
```
RemyDeWolf, line 11: we can use the object type like what we do with datadog or helm charts values
https://github.com/run-x/opta/blob/main/modules/datadog/datadog.json#L15-L18

modules/external_state/external-state.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$id": "https://app.runx.dev/modules/external-state",
+  "type": "object",
+  "description": "Adds the ability to refer to external, preexisting terraform states in opta.",
+  "properties": {
+    "backend_type": {
+      "type": "string",
+      "description": "The type of backend this external state is using"
```
RemyDeWolf, line 8: can we also add the link here https://www.terraform.io/language/settings/backends/configuration#backend-types

modules/external_state/external-state.md
```diff
@@ -0,0 +1,84 @@
...
+weight: 1
+description: Adds the ability to refer to external, preexisting terraform states in opta.
+---
+
+This module grants users the ability to import outputs from 
+[external terraform backends](https://www.terraform.io/language/settings/backends) (e.g. pre-existing
+or created outside of opta, or another Opta service/environment). All the terraform outputs of that state
+will be captured and referenceable under the `outputs` output for this module. 
+
+For instance, you can refer values such as the name of an S3 bucket created in another Opta service like so:
```
RemyDeWolf, line 15: I can see some possible confusion for some users between the S3 used for the backend and the S3 bucket used in this example. But they are different things.

Can we use another example? may be name of EKS cluster, Database name or VPC id these are usually common.

modules/external_state/external-state.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+The service will now have an environment variables known as SOURCE_BUCKET which could be the name of the
+bucket they wish grab data from (assuming IAM permissions are configured externally for simplicity).
+
+{{% alert title="NOTE" color="orange" %}}
+Usage is reliant on what the external state outputs and names. Please make sure that the data you seek
+is both outputted and has the expected name. If the state is created by Opta you can find the outputs
+and name by running opta apply on that state's manifest yaml.
+{{% /alert %}}
```
RemyDeWolf, line 47: what will happen if the output name is not found?

modules/external_state/external-state.md
```diff
@@ -0,0 +1,84 @@
+---
+title: "external-state"
+linkTitle: "external-state"
+date: 2021-07-21
+draft: false
+weight: 1
+description: Adds the ability to refer to external, preexisting terraform states in opta.
```
RemyDeWolf, line 7: random question, can this module be used when referring to a stack managed by opta?

I don't know if it's an interesting use case.

modules/external_state/tf_module/main.tf
```diff
@@ -0,0 +1,5 @@
+data "terraform_remote_state" "remote" {
+  backend = var.backend_type
+
+  config = var.config
+}
```
RemyDeWolf, line 5: 👍 woo, that was it!

modules/external_state/external-state.md
juandiegopalomino: `(e.g. pre-existing
or created outside of opta, or another Opta service/environment)`

RemyDeWolf (approved):

PR #769, opened 2022-03-14
Module api/validation

bigbitbus:
# Description
Added original validation (yamale based) to module API.

# Safety checklist

YOUR_ANSWER


rabbitfang:
Pausing module API work

PR #770, opened 2022-03-14
Fixing gke high/critical security issues

bigbitbus:
# Description
Fixing/ignoring high and critical tfsec red flags.

# Safety checklist

TBD - circleCI e2e tests


juandiegopalomino (approved):
And this does not force a cluster replacement, yes?

bigbitbus:
> And this does not force a cluster replacement, yes?

I am checking that on a pre-existing cluster, will remove the WIP label once its confirmed.

PR #771, opened 2022-03-14
Adding the image name plus tag as output for k8s service

juandiegopalomino:
# Description
Useful to refer to current deployment in other modules of the same service (e.g. jobs) as well as see for debugging

# Safety checklist

Manually


bigbitbus (approved):
good usability improvement.

PR #772, opened 2022-03-14
Add module data member to hold additional helm chart values

nsarupr:
# Description

# Safety checklist

YOUR_ANSWER


modules/aws_k8s_base/aws-k8s-base.json
```diff
@@ -61,6 +61,21 @@
...
     },
+    "cert_manager_values": {
+      "type": "object",
+      "description": "Certificate Manager helm chart additional values. [Available options](https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=values)",
+      "default": {}
+    },
+    "linkerd_values": {
+      "type": "object",
+      "description": "Linkerd helm chart additional values. [Available options](https://artifacthub.io/packages/helm/linkerd2/linkerd2/2.10.2?modal=values)",
+      "default": {}
```
RemyDeWolf, line 72: the default is actually what we have here 
https://github.com/run-x/opta/blob/main/modules/aws_k8s_base/tf_module/values-ha.yaml
we can document this - so it's clear to the user where this values are coming from

modules/azure_k8s_base/azure-k8s-base.json
```diff
@@ -23,6 +23,21 @@
       "description": "Additional configuration for nginx ingress. [Available options](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#configuration-options)",
       "default": {}
     },
+    "cert_manager_values": {
+      "type": "object",
+      "description": "Certificate Manager helm chart additional values. [Available options](https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=values)",
+      "default": {}
+    },
+    "linkerd_values": {
```
RemyDeWolf, line 31: To go further, instead of hard coding annotation in the pods, we should use `podAnnotations` in our `/modules/aws_k8s_base/tf_module/values-ha.yaml`.

These annotations would be added in the values file and removed from the code.
```
        config.linkerd.io/skip-outbound-ports: "8126,5432,3306"
        config.linkerd.io/proxy-cpu-limit: "0.1"
        config.linkerd.io/proxy-cpu-request: "0.05"
        config.linkerd.io/proxy-memory-limit: "20Mi"
        config.linkerd.io/proxy-memory-request: "10Mi"
```

modules/azure_k8s_base/azure-k8s-base.json
```diff
@@ -23,6 +23,21 @@
...
+      "type": "object",
+      "description": "Linkerd helm chart additional values. [Available options](https://artifacthub.io/packages/helm/linkerd2/linkerd2/2.10.2?modal=values)",
```
RemyDeWolf, line 33: this creates one more thing to maintain when we update the linkerd version. You can try to define the version in a constant - and pass that version as a terraform input variable.

not blocking for this PR though.

modules/azure_k8s_base/azure-k8s-base.json
```diff
@@ -23,6 +23,21 @@
...
+      "default": {}
+    },
+    "ingress_nginx_values": {
+      "type": "object",
+      "description": "Ingress Nginx helm chart additional values. [Available options](https://artifacthub.io/packages/helm/ingress-nginx/ingress-nginx/4.0.17?modal=values)",
```
RemyDeWolf, line 38: similar comment about defining a version constant to avoid hard coded the version in the url. 
not blocking.

modules/gcp_k8s_base/gcp-k8s-base.yaml
```diff
@@ -71,6 +71,21 @@ inputs:
     validator: bool(required=False)
     description: Expose self-signed SSL certs.
     default: false
+  - name: cert_manager_values
+    user_facing: true
+    validator: any(required=False)
+    description: Certificate Manager helm chart additional values. [Available options](https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=values)
+    default: {}
```
RemyDeWolf, line 78: I don't think we should change cert manager since we want to remove it https://run-x.atlassian.net/browse/RUNX-1216

if some users use this new values then we won't be able to remove it.

modules/aws_k8s_base/aws-k8s-base.json
nsarupr: The defaults for High availability we already have, but we are just appending any more that the user would want.

modules/azure_k8s_base/azure-k8s-base.json
nsarupr: These values are not just for HA linkerd, but for the default availability as well. 
Either we can make 2 values file, one with the high availability which we already have, and we can make one for the regular availability as well, so the users can then control them as well.

modules/azure_k8s_base/azure-k8s-base.json
nsarupr: I know this will create some additional works, as and when we upgrade the versions, but this enabled users to use the chart values of the specific version, and not make some unintentional mistakes.

modules/azure_k8s_base/azure-k8s-base.json
nsarupr: We want to test these things before giving these out to our users.

PR #774, opened 2022-03-16
TFsec - Remove egress from database, documentdb and elasticcache SGs

bigbitbus:
# Description

Databases don't need egress rules to 0.0.0.0/0 to all ports. Only ingress needed.

# Safety checklist

CircleCI tbd.
CircleCI (look at only the AWS cloud passing tests - https://app.circleci.com/pipelines/github/run-x/opta/2168/workflows/3a5e7976-8f35-4d82-8a53-f77aeb9ca111)

### DIFF
--- a/modules/aws_base/tf_module/log_bucket.tf
+++ b/modules/aws_base/tf_module/log_bucket.tf
@@ -13,6 +13,16 @@ resource "aws_s3_bucket" "log_bucket" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "log_bucket" {
+  bucket = aws_s3_bucket.log_bucket.id
+
+  block_public_acls   = true
+  block_public_policy = true
+  restrict_public_buckets = true
+  ignore_public_acls = true
+
+}
+
 resource "aws_s3_bucket_acl" "log_bucket" {
   bucket = aws_s3_bucket.log_bucket.id
   acl    = "log-delivery-write"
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -1552,7 +1552,7 @@ jobs:
       OPTA_DISABLE_REPORTING: "true"
     steps:
       - checkout-opta
-      - install-python
+      # - install-python
```
rabbitfang, line 1555:
```suggestion
```

If python isn't needed, the line should be removed entirely instead of commented out. Also, this change seems out of scope for this PR and could be done separately. (Ditto for the equivalent changes later in the file)

rabbitfang (commented):
Otherwise, LGTM

.circleci/config.yml
bigbitbus: This was removed - Nilesh did the circleci changes in another PR

bigbitbus:
@rabbitfang addressed your CR, got passing tests for AWS in CircleCI. Can you approve. Thanks!

modules/aws_base/tf_module/log_bucket.tf
```diff
@@ -13,6 +13,16 @@ resource "aws_s3_bucket" "log_bucket" {
   }
 }
 
+resource "aws_s3_bucket_public_access_block" "log_bucket" {
+  bucket = aws_s3_bucket.log_bucket.id
+
+  block_public_acls   = true
+  block_public_policy = true
+  restrict_public_buckets = true
+  ignore_public_acls = true
```
rabbitfang, line 22: Can you run `terraform fmt` on this file? These lines aren't aligned.

bigbitbus: Done and pushed

rabbitfang (approved):

PR #775, opened 2022-03-16
Feature/tfsec azure critical fixes

bigbitbus:
# Description
Plugging or ignoring various critical flagged tfsec Azure issues.

# Safety checklist

CircleCI Azure TBD.

### DIFF
--- a/modules/azure_base/tf_module/encryption.tf
+++ b/modules/azure_base/tf_module/encryption.tf
@@ -16,10 +16,10 @@ resource "azurerm_key_vault" "opta" {
   lifecycle {
     ignore_changes = [location]
   }
-  network_acls {
-    bypass         = "AzureServices"
-    default_action = "Deny"
-  }
+  # network_acls {
+  #   bypass         = "AzureServices"
+  #   default_action = "Deny"
+  # }
 }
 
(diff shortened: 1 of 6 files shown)


modules/azure_base/tf_module/encryption.tf
```diff
@@ -16,6 +16,10 @@ resource "azurerm_key_vault" "opta" {
   lifecycle {
     ignore_changes = [location]
   }
+  # network_acls {
+  #   bypass         = "AzureServices"
+  #   default_action = "Deny"
+  # }
```
rabbitfang, line 22: Should this section be removed?

modules/azure_k8s_base/tf_module/variables.tf
```diff
@@ -42,7 +42,9 @@ variable "linkerd_high_availability" {
 }
 
 variable "private_key" {
-  type    = string
+  type = string
+  # Ignore since this default value is never used, its for a logical if condition checking if user specified no private_key only.
+  #tfsec:ignore:general-secrets-no-plaintext-exposure
```
rabbitfang, line 47: Instead of ignoring the issue caused by the empty string, can we use a `null` value instead?

juandiegopalomino (approved):

PR #778, opened 2022-03-21
Use current latest version of External DNS helm chart

nsarupr:
# Description
Upgrade the External DNS Helm chart version used for AWS.

# Safety checklist

Tested Manually by upgrading from `3.7.0` to `6.2.1` and didn't see any downtime on the `Hello Opta` k8s-service.


RemyDeWolf (commented):
do you need to use the DNS module to test this change?

nsarupr:
> do you need to use the DNS module to test this change?

Either have a `domain` variable user defined or have a DNS module and `aws_k8s_base.py` would try to interpolate values.

RemyDeWolf (approved):

PR #783, opened 2022-03-22
Configurable values chart

nsarupr:
# Description
Add additional values to the Helm charts for Cert Manager, Linkerd and Ingress Nginx

# Safety checklist

Tested Manually


modules/aws_k8s_base/aws-k8s-base.json
```diff
@@ -61,6 +61,21 @@
       "description": "Expose self-signed SSL certs.",
       "default": false
     },
+    "cert_manager_values": {
+      "type": "object",
+      "description": "Certificate Manager helm chart additional values. [Available options](https://artifacthub.io/packages/helm/cert-manager/cert-manager?modal=values)",
+      "default": {}
+    },
```
RemyDeWolf, line 68: let's not change cert manager since we plan to delete it

RemyDeWolf (approved):
let's just not change cert manager but otherwise looks good

modules/aws_k8s_base/aws-k8s-base.json
nsarupr: I'll refactor this in another PR.

PR #784, opened 2022-03-22
Add user variable for adding custom name for Databases

nsarupr:
# Description
Now opta users can change the name of the databases with this new change.

> **Note:**
> **This change to some existing modules may lead to data losses, as terraform takes these changes as deleting old resources and creating new resources ones. So, for the already existing DBs, the names can't be changed.**

# Safety checklist

# Example
```yaml
environments:
  - name: aws
    path: "./aws.yml"
    variables:
      max_containers: 6
  - name: gcp
    path: "./gcp.yml"
    variables:
      max_containers: 6
input_variables:
  - name: name
name: test-naming
modules:
  - type: postgres
    name: postgres
    safety: false
    identifier: "{vars.name}"
  - name: redis
    type: redis
    identifier: "{vars.name}"
```

Manually tested.

Screenshot 2022-03-22 at 5 51 04 PM
Screenshot 2022-03-22 at 8 50 14 PM
Screenshot 2022-03-22 at 8 50 36 PM

### DIFF
--- a/modules/aws_documentdb/aws-documentdb.yaml
+++ b/modules/aws_documentdb/aws-documentdb.yaml
@@ -37,7 +37,7 @@ inputs:
   - name: identifier
     user_facing: true
     validator: str(required=False)
-    description: Name for the documentdb cluster.  If not specified, it will be look like opta-<layer_name>-<module_name>-<random_string>
+    description: Name for the documentdb cluster.  If not specified, it will default to opta-<layer_name>-<module_name>-<random_string>
     default: None
 outputs:
   - name: db_host
--- a/modules/aws_documentdb/aws-documentdb.yaml
+++ b/modules/aws_documentdb/aws-documentdb.yaml
@@ -38,7 +38,7 @@ inputs:
     user_facing: true
     validator: str(required=False)
     description: Name for the documentdb cluster.  If not specified, it will default to opta-<layer_name>-<module_name>-<random_string>
-    default: None
+    default: null
 outputs:
   - name: db_host
     export: true
--- a/modules/aws_documentdb/aws-documentdb.yaml
+++ b/modules/aws_documentdb/aws-documentdb.yaml
@@ -34,6 +34,11 @@ inputs:
     validator: bool(required=False)
     description: A value that indicates whether the DB cluster has deletion protection enabled. The database can't be deleted when deletion protection is enabled. By default, deletion protection is disabled.
     default: false
+  - name: identifier
+    user_facing: true
+    validator: str(required=False)
+    description: Name for the documentdb cluster.  If not specified, it will be look like opta-<layer_name>-<module_name>-<random_string>
+    default: None
 outputs:
   - name: db_host
     export: true
--- a/modules/aws_mysql/aws-mysql.json
+++ b/modules/aws_mysql/aws-mysql.json
@@ -29,6 +29,11 @@
       "description": "Add deletion protection to stop accidental db deletions",
       "default": false
     },
+    "identifier": {
+      "type": "string",
+      "description": "Identifier for the RDS instance.",
+      "default": null
+    },
     "type": {
       "description": "The name of this module",
       "enum": [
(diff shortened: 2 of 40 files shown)


modules/aws_mysql/aws-mysql.json
```diff
@@ -29,6 +29,10 @@
       "description": "Add deletion protection to stop accidental db deletions",
       "default": false
     },
+    "identifier": {
+      "type": "string",
+      "description": "Identifier for the RDS instance."
```
RemyDeWolf, line 34: let's say `The name of the RDS instance`
from https://registry.terraform.io/modules/terraform-aws-modules/rds/aws/latest#input_identifier

modules/aws_mysql/aws-mysql.yaml
```diff
@@ -39,6 +39,11 @@ inputs:
     validator: bool(required=False)
     description: Add deletion protection to stop accidental db deletions
     default: false
+  - name: identifier
+    user_facing: true
+    validator: str(required=False)
+    description: Identifier for the RDS instance.
```
RemyDeWolf, line 45: let's add something to explain the default value.
_If not set, use a generated unique name based on the layer and module name_

modules/aws_mysql/tf_module/variables.tf
```diff
@@ -37,3 +37,8 @@ variable "multi_az" {
   type    = bool
   default = false
 }
+
+variable "identifier" {
+  type    = string
+  default = null
```
RemyDeWolf, line 43: let's not add `default = null` since we would always set it, similar to env_name

modules/aws_redis/aws-redis.json
```diff
@@ -86,6 +86,10 @@
       "description": "Window during which the snapshot would be taken.",
       "default": "04:00-05:00"
     },
+    "identifier": {
+      "type": "string",
+      "description": "Identifier for the Redis instance."
```
RemyDeWolf, line 91: _The name of the Redis instance_

modules/aws_mysql/tf_module/variables.tf
nsarupr: I tried setting this to a default value which we use, but the interpolation is not allowed for variables in terraform and we do need this as `null` in order to check what name to set.

rabbitfang (changes requested):

modules/aws_documentdb/aws-documentdb.yaml
```diff
@@ -34,6 +34,11 @@ inputs:
     validator: bool(required=False)
     description: A value that indicates whether the DB cluster has deletion protection enabled. The database can't be deleted when deletion protection is enabled. By default, deletion protection is disabled.
     default: false
+  - name: identifier
```
bigbitbus, line 37: For all these identifier strings, do we need input validation?

For example, forbidden names of clusters, too long/too short, special characters not allowed etc. There may be some constraints on these...not

bigbitbus (commented):
What would happen if someone has a pre-existing database (any of the ones you have addressed in this code). This is currently named opta-layer-module....; now the user sees this new feature and adds identifier to their opta yaml file.

What is the behavior then? Will their pre-existing DB be renamed or deleted and replaced, for example?

modules/aws_documentdb/aws-documentdb.yaml
nsarupr: I'm not sure whether we should put validations or not on these.
Currently I left the validation to individual terraform or provider restrictions.

nsarupr:
> What would happen if someone has a pre-existing database (any of the ones you have addressed in this code). This is currently named opta-layer-module....; now the user sees this new feature and adds identifier to their opta yaml file.
> 
> What is the behavior then? Will their pre-existing DB be renamed or deleted and replaced, for example?

I tested this but in order to be totally sure, let me verify this thing once again.
At places these changes are being ignored in lifecycle...at some places they aren't.

PR #787, opened 2022-03-24
Fix value check for CORS for AWS S3 Bucket (#786)

nsarupr:
# Description
Fix the AWS Terraform Resource `aws_s3_bucket_cors_configuration` usage 

# Safety checklist

Tested manually
Screenshot 2022-03-25 at 12 18 10 AM


bigbitbus (commented):
How could we have avoided this before releasing 0.28.0?

Do we know how to write a test for this in CI?

nsarupr:
> How could we have avoided this before releasing 0.28.0?
> 
> Do we know how to write a test for this in CI?

In order to avoid this thing, we would need to exercise all the inputs available to us for S3 module.

rabbitfang (approved):

PR #788, opened 2022-03-25
Check for Helm Status in k8s service prehook

nsarupr:
# Description

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/base.py
+++ b/modules/base.py
@@ -197,14 +197,14 @@ class K8sServiceModuleProcessor(ModuleProcessor):
 
     def pre_hook(self, module_idx: int) -> None:
         pending_upgrade_helm_chart = Helm.get_helm_list(
-            release=f"{self.layer.name}-{self.module.name}"
-        )  # , status="pending-upgrade")
+            release=f"{self.layer.name}-{self.module.name}", status="pending-upgrade"
+        )
         if pending_upgrade_helm_chart:
             raise UserErrors(
                 f"There is a pending upgrade for the helm chart: {self.layer.name}-{self.module.name}."
                 "\nPlease use `opta force-unlock` to rollback already existing module."
             )
-        return super(K8sServiceModuleProcessor, self).pre_hook(module_idx)
+        return super().pre_hook(module_idx)
 
     @property
     def required_path_dependencies(self) -> FrozenSet[str]:
@@ -462,10 +462,6 @@ class K8sBaseModuleProcessor:
 
 
 class AWSK8sModuleProcessor(ModuleProcessor):
-    def pre_hook(self, module_idx: int) -> None:
-        print("AWSK8sModuleProcessor.pre_hook")
-        return super(AWSK8sModuleProcessor, self).pre_hook(module_idx)
-
     def __init__(self, module: "Module", layer: "Layer"):
         super(AWSK8sModuleProcessor, self).__init__(module, layer)
 
--- a/modules/base.py
+++ b/modules/base.py
@@ -196,12 +196,13 @@ class K8sServiceModuleProcessor(ModuleProcessor):
     FLAG_MULTIPLE_PORTS_SUPPORTED = False
 
     def pre_hook(self, module_idx: int) -> None:
+        release_name = f"{self.layer.name}-{self.module.name}"
         pending_upgrade_helm_chart = Helm.get_helm_list(
-            release=f"{self.layer.name}-{self.module.name}", status="pending-upgrade"
+            release=release_name, status="pending-upgrade"
         )
         if pending_upgrade_helm_chart:
             raise UserErrors(
-                f"There is a pending upgrade for the helm chart: {self.layer.name}-{self.module.name}."
+                f"There is a pending upgrade for the helm chart: {release_name}."
                 "\nIt will cause this command to fail. Please use `opta force-unlock` to rollback to a consistent state first."
             )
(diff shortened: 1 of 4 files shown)


modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
     # TODO(patrick): Remove this flag and references to it once all clouds support multiple ports
     FLAG_MULTIPLE_PORTS_SUPPORTED = False
 
+    def pre_hook(self, module_idx: int) -> None:
+        pending_upgrade_helm_chart = Helm.get_helm_list(
+            release=f"{self.layer.name}-{self.module.name}"
+        )  # , status="pending-upgrade")
```
rabbitfang, line 201: The status check is disabled here which does not appear to be intended behavior, since that would cause the below `if` statement to always be `True` if the chart exists, even if installed correctly.

modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
...
+        if pending_upgrade_helm_chart:
+            raise UserErrors(
+                f"There is a pending upgrade for the helm chart: {self.layer.name}-{self.module.name}."
+                "\nPlease use `opta force-unlock` to rollback already existing module."
+            )
+        return super(K8sServiceModuleProcessor, self).pre_hook(module_idx)
```
rabbitfang, line 207:
```suggestion
        return super().pre_hook(module_idx)
```

Use of the short form of `super` is preferred in Python 3.

modules/base.py
```diff
@@ -450,6 +462,10 @@ def __process_nginx_extra_ports(
 
 
 class AWSK8sModuleProcessor(ModuleProcessor):
+    def pre_hook(self, module_idx: int) -> None:
+        print("AWSK8sModuleProcessor.pre_hook")
+        return super(AWSK8sModuleProcessor, self).pre_hook(module_idx)
+
```
rabbitfang, line 468:
```suggestion
```

This code looks like it was left in from debugging

rabbitfang (dismissed):

modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
     # TODO(patrick): Remove this flag and references to it once all clouds support multiple ports
     FLAG_MULTIPLE_PORTS_SUPPORTED = False
 
+    def pre_hook(self, module_idx: int) -> None:
```
rabbitfang, line 198: This method should also have test coverage

modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
...
+        pending_upgrade_helm_chart = Helm.get_helm_list(
+            release=f"{self.layer.name}-{self.module.name}", status="pending-upgrade"
+        )
+        if pending_upgrade_helm_chart:
+            raise UserErrors(
+                f"There is a pending upgrade for the helm chart: {self.layer.name}-{self.module.name}."
+                "\nPlease use `opta force-unlock` to rollback already existing module."
```
NitinAgg, line 205:
```suggestion
                "\n It will cause this command to fail. Please use `opta force-unlock` to rollback to a consistent state first"
```

modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
     # TODO(patrick): Remove this flag and references to it once all clouds support multiple ports
     FLAG_MULTIPLE_PORTS_SUPPORTED = False
 
+    def pre_hook(self, module_idx: int) -> None:
+        pending_upgrade_helm_chart = Helm.get_helm_list(
+            release=f"{self.layer.name}-{self.module.name}", status="pending-upgrade"
```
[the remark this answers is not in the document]
nsarupr: It won't. They are just filters.

modules/base.py
```diff
@@ -194,6 +195,17 @@ class K8sServiceModuleProcessor(ModuleProcessor):
...
+        )
+        if pending_upgrade_helm_chart:
+            raise UserErrors(
+                f"There is a pending upgrade for the helm chart: {self.layer.name}-{self.module.name}."
```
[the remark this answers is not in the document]
nsarupr: Just creates one more variable which won't be used anywhere else.

modules/base.py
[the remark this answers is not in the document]
RemyDeWolf: ok let's add a docstring or comment in the `get_helm_list` so we will remember this info

modules/base.py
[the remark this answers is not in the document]
RemyDeWolf: it is used twice in this bloc

opta/core/helm.py
```diff
@@ -40,6 +40,10 @@ def rollback_helm(cls, release: str, namespace: str, revision: str = "") -> None
 
     @classmethod
     def get_helm_list(cls, **kwargs) -> List:  # type: ignore # nosec
+        """
+        Returns a list of helm releases.
+        The releases can be filtered by namespace, release name and status.
+        """
```
RemyDeWolf, line 46: 👍

RemyDeWolf (approved):

opta/core/helm.py
RemyDeWolf: also say something that no exception is thrown if not found

PR #791, opened 2022-03-29
Use aws-eks instead of k8s-cluster

nsarupr:
# Description
Use the actual type instead of aliased type.

# Safety checklist

YOUR_ANSWER


bigbitbus (commented):
lgtm, have you had a chance to test it?

nsarupr:
> lgtm, have you had a chance to test it?

Worked well on an already existing environment. Currently testing by spawning a fresh environment and checking a few things in that.

NitinAgg (approved):

juandiegopalomino (approved):

RemyDeWolf (commented):
why do we need this change? in the docs this module goes by `k8s-cluster` https://docs.opta.dev/reference/aws/modules/aws-eks/

PR #793, opened 2022-03-30
Add input variable db_name to aws-mysql and aws-postgres

nsarupr:
# Description
This let's users add a custom database name created with `aws-mysql` and `aws-postgres` or `mysql` and `postgres` modules respectively for AWS provider

> **Note:**
> **If the db_name is changed, the terraform resources will be destroyed and created again.**

# Safety checklist

Tested Manually by creating a MySQL module.

## Observations:
The current existing module resources will be destroyed and new module resources would be created, if the users change the `db_name`

### DIFF
--- a/modules/aws_mysql/aws-mysql.json
+++ b/modules/aws_mysql/aws-mysql.json
@@ -31,7 +31,7 @@
     },
     "db_name": {
       "type": "string",
-      "description": "The name of the database to create.",
+      "description": "The name of the database to create. Follow naming conventions [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html#RDS_Limits.Constraints)",
       "default": "app"
     },
     "type": {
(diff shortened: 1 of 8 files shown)


modules/aws_mysql/aws-mysql.json
```diff
@@ -29,6 +29,11 @@
       "description": "Add deletion protection to stop accidental db deletions",
       "default": false
     },
+    "db_name": {
+      "type": "string",
+      "description": "The name of the database to create.",
```
RemyDeWolf, line 34: looks good, the official documentation mentions some restrictions, may be we could also include them -> https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster#database_name

nsarupr: If the official documentation already covers this, it would just be a redundancy on our part to do this thing again.

RemyDeWolf: the opta users are not exposed to the official terraform documentation, they only see https://docs.opta.dev/reference/aws/modules/aws-mysql/
or did you mean something else?

RemyDeWolf (approved):
approving, pending the documentation change that we discussed

RemyDeWolf (approved):

PR #799, opened 2022-04-01
Support for multi-region postgres dbs on aws

juandiegopalomino:
# Description
See opta docs pr for full details
https://github.com/run-x/opta-docs/pull/173

# Safety checklist

manually


modules/aws_postgres/aws-postgres.yaml
```diff
@@ -39,10 +39,25 @@ inputs:
     validator: int(required=False)
     description: How many days to keep the backup retention
     default: 7
-  - name: db_name
+  - name: extra_security_groups
+    user_facing: true
+    validator: list(str(), required=False)
+    description: Extra AWS security groups to add to the database
```
RemyDeWolf, line 45: is it a name or id?

modules/aws_postgres/aws-postgres.yaml
```diff
@@ -39,10 +39,25 @@ inputs:
...
+  - name: existing_global_database_id
+    user_facing: true
+    validator: str(required=False)
+    description: ID of the Aurora global database to attach
+    default: null
+  - name: database_name
     user_facing: true
     validator: str(required=False)
-    description: The name of the database to create. Follow naming conventions [here](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_Limits.html#RDS_Limits.Constraints)
+    description: Name of the database to create within the instance
```
RemyDeWolf, line 60: why removing the official naming conventions? our users can run into this issue and create support for us otherwise

modules/aws_postgres/tf_module/outputs.tf
```diff
@@ -1,20 +1,25 @@
 output "db_user" {
-  value      = aws_rds_cluster.db_cluster.master_username
-  depends_on = [aws_rds_cluster_instance.db_instance[0]]
+  value      = var.existing_global_database_id == null ? aws_rds_cluster.db_cluster[0].master_username : "UNKNOWN_SEE_MASTER_DB"
```
RemyDeWolf, line 2: we can also use PRIMARY here

modules/aws_postgres/tf_module/main.tf
```diff
@@ -18,37 +18,87 @@ resource "random_string" "db_name_hash" {
...
   apply_immediately               = true
   auto_minor_version_upgrade      = false
   performance_insights_enabled    = true
   performance_insights_kms_key_id = data.aws_kms_key.main.arn
   lifecycle {
     ignore_changes = [identifier]
   }
+}
+
+resource "aws_rds_global_cluster" "global_cluster" {
```
RemyDeWolf, line 57: suggestion: is there any way we can break down this file into smaller tf files?
may be: `rds.tf` and `rds_global` that way if someone use the tf generator they could just remove the file that they don't need.

RemyDeWolf (approved):

PR #800, opened 2022-04-04
Give Opta users ability to add and update pod annotations for the K8s-service pod objects

nsarupr:
# Description
This feature will let the Opta Users, existing and new, **add new and update existing pod annotations** in the Pod resources created using the opta module `k8s-service/aws-k8s-service/azure-k8s-service/gcp-k8s-service`

# Safety checklist

Screenshot 2022-04-04 at 8 34 42 PM
Screenshot 2022-04-04 at 8 34 24 PM
Screenshot 2022-04-04 at 8 18 44 PM
Screenshot 2022-04-04 at 8 19 30 PM
Screenshot 2022-04-04 at 8 19 02 PM
Screenshot 2022-04-04 at 8 19 47 PM

### DIFF
--- a/modules/aws_k8s_service/aws-k8s-service.json
+++ b/modules/aws_k8s_service/aws-k8s-service.json
@@ -175,6 +175,11 @@
       "description": "These are extra annotations to add to ingress objects\n",
       "default": {}
     },
+    "pod_annotations": {
+      "type": "object",
+      "description": "These are extra annotations to add to k8s-service pod objects \n",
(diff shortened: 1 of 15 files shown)


juandiegopalomino (changes requested):
move the current annotations out of helm into terraform as a dict and then merge that "default" dict with the one which is user provided (so that users can reset ones we already have)

juandiegopalomino (approved):

PR #801, opened 2022-04-04
Opta shell needs to load in the kubeconfig for it to work

juandiegopalomino:
# Description
Fails with a weird message if there was another environment loaded up as default

# Safety checklist

Manually

### DIFF
--- a/opta/core/aws.py
+++ b/opta/core/aws.py
@@ -435,16 +435,21 @@ class AWS(CloudClient):
         except client.exceptions.ResourceNotFoundException:
             return False
 
+    def get_kubectl_context_name(self) -> str:
+        region, account_id = self.get_cluster_env()
+        # Get the environment's account details from the opta config
+        cluster_name = self.layer.get_cluster_name()
+        return f"{account_id}_{region}_{cluster_name}"
+
     def set_kube_config(self) -> None:
         kube_config_file_name = self.layer.get_kube_config_file_name()
+        region, account_id = self.get_cluster_env()
         if exists(kube_config_file_name):
             if getmtime(kube_config_file_name) > time() - ONE_WEEK_UNIX:
                 constants.GENERATED_KUBE_CONFIG = kube_config_file_name
                 return
             else:
                 remove(kube_config_file_name)
-
-        region, account_id = self.get_cluster_env()
 
         # Get the environment's account details from the opta config
         cluster_name = self.layer.get_cluster_name()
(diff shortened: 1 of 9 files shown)


NitinAgg (approved):

opta/core/cloud_client.py
```diff
@@ -41,3 +41,7 @@ def cluster_exist(self) -> bool:
     @abstractmethod
     def set_kube_config(self) -> None:
         raise NotImplementedError()
+
+    @abstractmethod
+    def get_kubectl_context_name(self) -> str:
+        pass
```
RemyDeWolf, line 47: why not raise NotImplementedError() like others?

opta/core/aws.py
```diff
@@ -435,17 +435,22 @@ def cluster_exist(self) -> bool:
         except client.exceptions.ResourceNotFoundException:
             return False
 
+    def get_kubectl_context_name(self) -> str:
```
RemyDeWolf, line 438: for the name, the kubeconfig is not only specific to kubectl, so we can just call it `kube_context`
example of another tool using it https://skaffold.dev/docs/environment/kube-context/

opta/core/aws.py
```diff
@@ -435,17 +435,22 @@ def cluster_exist(self) -> bool:
...
+        region, account_id = self.get_cluster_env()
```
RemyDeWolf, line 439: you can have a region and account without a cluster - may be we we should also rename it `get_provider_data` or `get_cloud_env` or something that is not tied to a cluster.

I realized this is not a new method - but since we use it, we can rename it if we wish.

opta/core/gcp.py
```diff
@@ -222,6 +222,12 @@ def get_cluster_env(self) -> Tuple[str, str]:
         googl_provider = self.layer.root().providers["google"]
         return googl_provider["region"], googl_provider["project"]
 
+    def get_kubectl_context_name(self) -> str:
```
RemyDeWolf, line 225: this is a nice abstraction - anyway we can add a unit test just on this method? `test_kubectl_context_name` (aws/gcp)

opta/core/cloud_client.py
juandiegopalomino: it actually does if you try to run

opta/core/gcp.py
juandiegopalomino: it's implementations are so small, that not really without mocking most of it

opta/core/aws.py
juandiegopalomino: nah, let's keep it as is for now

PR #802, opened 2022-04-04
Add support for bring-your-own-vpc to aws-base

rabbitfang:
# Description
- Add vpc_id, public_subnet_ids, and private_subnet_ids fields to aws-base
- When set, will not create our own VPC but instead use the one provided
- Does not set up flow logs when using an existing VPC

# Safety checklist

- No change in terraform plan when using this code vs `main` branch when not using BYO-VPC.
- Manually created a VPC in the AWS console and tested using that VPC, verified that VPC was used when creating an EKS cluster via opta

### DIFF
--- a/modules/aws_base/aws-base.md
+++ b/modules/aws_base/aws-base.md
@@ -9,3 +9,14 @@ description: Sets up VPCs, a default KMS key, and the db/cache subnets for your 
 
 The defaults for this module are set to work 99% of the time, assuming no funny networking constraints (you'll know them
 if you have them), so in most cases, there is _no need to set any of the fields or know what the outputs do_.
+
+## Bring your own VPC
+To use an existing VPC with Opta, instead of having Opta create a new VPC, you must set the `vpc_id`, `public_subnet_ids`, and `private_subnet_ids` fields.
+Set `vpc_id` to the ID of the existing VPC you would like Opta to use.
+Set `public_subnet_ids` to the list of IDs of the public subnets in the VPC.
+Public subnets must have a route that connects to an internet gateway, and must be configured to assign public IP addresses.
+Set `private_subnet_ids` to the list of IDs of the private subnets in the VPC.
+Private subnets must have a route with a destination of `0.0.0.0/0` that points to a NAT gateway with a public IP address.
+
+IPv6 is not supported on VPCs that you bring.
+Those VPCs may work, but we do not verify Opta works properly with IPv6-only or dual-stack VPCs.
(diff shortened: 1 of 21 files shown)


modules/aws_base/tf_module/existing_vpc.tf
```diff
@@ -0,0 +1,37 @@
...
+  count = local.create_vpc ? 0 : 1
+  id    = var.vpc_id
+  state = "available"
+}
+
+data "aws_subnet" "public_subnets" {
+  for_each = local.create_vpc ? toset([]) : toset(var.public_subnet_ids)
+  id       = each.key
+  vpc_id   = var.vpc_id
+  # state    = "available"
```
RemyDeWolf, line 11: what was the story behind uncommenting the state for the various imported resources?
shall they be removed?

modules/aws_base/aws-base.md
```diff
@@ -9,3 +9,14 @@ description: Sets up VPCs, a default KMS key, and the db/cache subnets for your
...
 The defaults for this module are set to work 99% of the time, assuming no funny networking constraints (you'll know them
 if you have them), so in most cases, there is _no need to set any of the fields or know what the outputs do_.
+
+## Bring your own VPC
+To use an existing VPC with Opta, instead of having Opta create a new VPC, you must set the `vpc_id`, `public_subnet_ids`, and `private_subnet_ids` fields.
+Set `vpc_id` to the ID of the existing VPC you would like Opta to use.
+Set `public_subnet_ids` to the list of IDs of the public subnets in the VPC.
+Public subnets must have a route that connects to an internet gateway, and must be configured to assign public IP addresses.
+Set `private_subnet_ids` to the list of IDs of the private subnets in the VPC.
+Private subnets must have a route with a destination of `0.0.0.0/0` that points to a NAT gateway with a public IP address.
```
RemyDeWolf, line 19: suggestion: For the must-have listed here, what would happen if they don't have them?
would terraform fail in a way that inform the user about what's wrong?
we could also consider a pre hook to check for the requirements so we can provider a user with a clear error message if the terraform output will not be enough. 

if it's too complicated to do such thing, I think we can skip it, this documentation is pretty clear.

RemyDeWolf (commented):
If I understand well, this would work with any VPC, even if they have not used terraform to provision them, we just use terraform data source to import them? this is brilliant!

I have a few minor comments.
would also be great to have a CI test for this feature, but we can do it later in a different PR.

modules/aws_base/tf_module/existing_vpc.tf
juandiegopalomino: ^

juandiegopalomino (approved):
lgtm, except for the state = "available" question

modules/aws_base/aws-base.md
RemyDeWolf: Ok, may be we can just _document_ what to do if the user gets this error message, to save us opta support on it.

RemyDeWolf (approved):

PR #807, opened 2022-04-07
Global Accelerator and the reformed cloudfront experience

juandiegopalomino:
# Description
OK so this is a long one, but here's what's happening:

In Opta prior to this change, we were not handling dns routing to none-k8s-base well. Like if you had cloudfront with k8s-base and the dns module, Opta couldn't point the traffic to cloudfront, and moreover if someone wrote the records themselves, they would be overwritten by external-dns.

As I was writing the global accelerator, I realized that I was going to face the same problems and had to take a holistic approach in stream lining this process. tl;dr read the modified documents, cause now not only is it super straightforward, but along the way I gave first class support to manual overrides so that Opta can easily integrate with pre-existing dns even better.

Moreover, we are deprecating external-dns as now thanks to the AWS load balancer controller we can just maintain control of our route53 records within opta no problem

Oh, and global accelerator works as you'd think, tried it out my self and with nilesh, and worked great.

# Safety checklist

manually test extensively


RemyDeWolf (commented):
let's figure out if we should expose `enable_auto_dns`  in each relevant module or have a new field `module_root` in the dns module itself.

modules/aws_dns/aws-dns.yaml
```diff
@@ -42,6 +42,11 @@ inputs:
     user_facing: false
     description: Deprecated
     default: ""
+  - name: linked_module
+    user_facing: true
+    validator: str(required=False)
+    description: The module type (or name if given) to automatically add root dns records for.
+    default: ""
```
NitinAgg, line 49: Let's make default `k8s-base` instead of handling None in the code

juandiegopalomino: Valid point, but these defaults aren't exposed in the code-- they're only filled in once terraform generation is done.

NitinAgg: oh, ok.

modules/aws_k8s_base/aws_k8s_base.py
```diff
@@ -43,16 +40,32 @@ def process(self, module_idx: int) -> None:
...
+                self.module.data.get("cert_arn")
+                or f"${{{{module.{aws_dns_module.name}.cert_arn}}}}"
+            )
+            self.module.data["zone_id"] = (
+                self.module.data.get("zone_id")
+                or f"${{{{module.{aws_dns_module.name}.zone_id}}}}"
+            )
+
+        if (self.module.data.get("domain") is None) != (
+            self.module.data.get("zone_id") is None
```
nsarupr, line 66: Default value for `domain` and `zone_id` is an empty string...might cause unexpected behaviour

modules/aws_k8s_base/tf_module/variables.tf
```diff
@@ -15,6 +15,7 @@ locals {
     ssl-redirect : true
     force-ssl-redirect : true
   }), var.nginx_config)
+  load_balancer_name = "opta-${substr(var.layer_name, 0, 22)}-lb"
```
nsarupr, line 18: any reasons why we are using substrings?
And better if we need such a limit, it should go directly on the `layer_name`

modules/cloudfront_distribution/cloudfront-distribution.md
```diff
@@ -22,14 +22,13 @@ providers:
...
+    domain: staging.startup.com
+    delegated: false # Set to true when ready -- see the "Configure DNS" page
+    linked_module: cloudfront-distribution
   - type: cloudfront-distribution
-    # Uncomment the following and fill in to support your domain with ssl
-#    acm_cert_arn: "arn:aws:acm:us-east-1:XXXXXXXXXX:certificate/cert-id"
-#    domains:
-#      - "your.domain.com"
     links:
       - testmodule
```
nsarupr, line 33: Documentation refactor required.

modules/cloudfront_distribution/cloudfront_distribution.py
```diff
@@ -39,6 +39,25 @@ def process(self, module_idx: int) -> None:
                 "s3_log_bucket_name", f"${{{{{module_source}.s3_log_bucket_name}}}}"
             )
 
+        aws_dns_modules = self.layer.get_module_by_type("aws-dns", module_idx)
+        if len(aws_dns_modules) != 0 and aws_dns_modules[0].data.get("linked_module") in [
+            self.module.type,
+            self.module.name,
```
nsarupr, line 45: Should there be a restriction on the what linked_modules can accept?
Or rather some sort of warning or error log, otherwise the user might just add anything, and not know whether it is supposed to work or not.

modules/aws_k8s_base/aws_k8s_base.py
juandiegopalomino: they're only filled in once terraform generation is done, so they are None

modules/aws_k8s_base/tf_module/variables.tf
juandiegopalomino: because I got errors that the lb name was too long. I can't change to a new name cause that would break everybody, so I opted out for this one, which will keep everyone's name as is (because if they're longer than 22 characters they would have failed)

nsarupr: Then this is still a breaking change for existing users, as this could potentially change the Load balancer url for existing users.
Need to mention this in the Release Docs as such.

RemyDeWolf (approved):

modules/cloudfront_distribution/cloudfront_distribution.py
juandiegopalomino: @nsarupr added

nsarupr (approved):

PR #812, opened 2022-04-11
Feature (AWS): ami_type in k8s-cluster/aws-eks and aws-nodegroup modules

nsarupr:
# Description
New Feature: Give users the capability to choose the AMI type for the nodegroups in the cluster as well as additional nodegroups.

> Note: The already existing nodegroups and clusters will not be updated and will need to be recreated, if one wishes to change the AMI Type

> Note: For users who are using **AMI type: AL2_x86_64_GPU** using `use_gpu: true` input in **aws-nodegroup** can now switch to using `ami_type: AL2_x86_64_GPU` as well. 
> This will just edit the labels, i.e. `gpu : "true"` -> `ami_type : AL2_x86_64_GPU`
> Before:
> ```yaml
>   - type: aws-nodegroup
>     name: nodegroup1
>     spot_instances: true
>     use_gpu: true
>     min_nodes: 1
>     max_nodes: 3
> ```
> After
> ```yaml
>   - type: aws-nodegroup
>     name: nodegroup1
>     spot_instances: true
>     ami_type: "AL2_x86_64_GPU"
>     min_nodes: 1
>     max_nodes: 3
> ```

# Safety checklist

Tested Manually with the above changes to make sure this is backwards compatible.

### DIFF
--- a/modules/aws_eks/aws-eks.json
+++ b/modules/aws_eks/aws-eks.json
@@ -63,6 +63,11 @@
       "default": {},
       "description": "Custom launch template for the underlying ec2s."
     },
+    "ami_type": {
+      "type": "boolean",
+      "description": "The AMI type to use for the nodes.\nFor more information about this, please visit [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType)\nNote: Currently, \"CUSTOM\" ami type is not supported.\n",
+      "default": "AL2_x86_64"
+    },
     "type": {
       "description": "The name of this module",
       "enum": [
--- a/modules/aws_nodegroup/aws-nodegroup.json
+++ b/modules/aws_nodegroup/aws-nodegroup.json
@@ -33,10 +33,10 @@
       "default": "t3.medium",
       "description": "The [ec2 instance type](https://aws.amazon.com/ec2/instance-types/) for the nodes."
     },
-    "use_gpu": {
+    "ami_type": {
       "type": "boolean",
-      "description": "Should we expect and use the gpus present in the ec2?",
-      "default": false
+      "description": "The AMI type to use for the nodes.\nFor more information about this, please visit [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType)\nNote: Currently, \"CUSTOM\" ami type is not supported.\n",
+      "default": "AL2_x86_64"
     },
     "spot_instances": {
       "type": "boolean",
--- a/modules/aws_nodegroup/aws-nodegroup.json
+++ b/modules/aws_nodegroup/aws-nodegroup.json
@@ -38,6 +38,11 @@
       "description": "The AMI type to use for the nodes.\nFor more information about this, please visit [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType)\nNote: Currently, \"CUSTOM\" ami type is not supported.\n",
       "default": "AL2_x86_64"
     },
+    "use_gpu": {
+      "type": "boolean",
+      "description": "Should we expect and use the gpus present in the ec2?\nNote: This input would be deprecated in the coming releases. Please switch to using `ami_type`\n",
+      "default": false
+    },
     "spot_instances": {
       "type": "boolean",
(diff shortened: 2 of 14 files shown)


juandiegopalomino (approved):
It looks good, but I worry if it will be too confusing for users who might just want something as simple as "Vanilla", "GPU", and "BottleRocket" as the "ami_type"-- prefer not to have them need to read the manual (because nobody likes doing that). This is just a personal preference and if @NitinAgg or @RemyDeWolf thinks this is fine, then we're good.

nsarupr:
> It looks good, but I worry if it will be too confusing for users who might just want something as simple as "Vanilla", "GPU", and "BottleRocket" as the "ami_type"-- prefer not to have them need to read the manual (because nobody likes doing that). This is just a personal preference and if @NitinAgg or @RemyDeWolf thinks this is fine, then we're good.

The request is genuine enough @juandiegopalomino but considering that we are using the actual enums for other places as well, such as when defining the `size` of a DB or when defining the which sort of cluster to use `RAPID/REGULAR` in GCP Cluster, I thought of using this.

modules/aws_eks/aws-eks.json
```diff
@@ -63,6 +63,11 @@
       "default": {},
       "description": "Custom launch template for the underlying ec2s."
     },
+    "ami_type": {
+      "type": "boolean",
+      "description": "The AMI type to use for the nodes.\nFor more information about this, please visit [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType)\nNote: Currently, \"CUSTOM\" ami type is not supported.\n",
```
RemyDeWolf, line 68: minor: let's keep custom lowercase

RemyDeWolf: should we direct them over this page instead?
https://docs.aws.amazon.com/eks/latest/userguide/eks-optimized-amis.html
it has the Bottlerocket AMIs too

modules/aws_nodegroup/aws-nodegroup.json
```diff
@@ -33,9 +33,14 @@
...
     },
+    "ami_type": {
+      "type": "boolean",
+      "description": "The AMI type to use for the nodes.\nFor more information about this, please visit [here](https://docs.aws.amazon.com/eks/latest/APIReference/API_Nodegroup.html#AmazonEKS-Type-Nodegroup-amiType)\nNote: Currently, \"CUSTOM\" ami type is not supported.\n",
+      "default": "AL2_x86_64"
+    },
     "use_gpu": {
       "type": "boolean",
-      "description": "Should we expect and use the gpus present in the ec2?",
+      "description": "Should we expect and use the gpus present in the ec2?\nNote: This input would be deprecated in the coming releases. Please switch to using `ami_type`\n",
```
RemyDeWolf, line 43: Shoudl we say, which value they should use in ami_type? `AL2_x86_64_GPU` right

modules/aws_nodegroup/aws_nodegroup.py
```diff
@@ -0,0 +1,20 @@
...
+    def __init__(self, module: "Module", layer: "Layer"):
+        super().__init__(module, layer)
+
+    def process(self, module_idx: int) -> None:
+        self.__back_compat_use_gpu()
+        super(AwsNodegroup, self).process(module_idx)
+
+    def __back_compat_use_gpu(self) -> None:
+        if self.module.data.get("use_gpu", False):
+            self.module.data["ami_type"] = "AL2_x86_64_GPU"
```
RemyDeWolf, line 20: should we display the deprecation warning here? only if it's only 1 message to the user, so they will address it faster.

modules/aws_eks/aws-eks.json
nsarupr: Too much information and complexity for the User in just one go. If the users want, they can research the supported AMI types on their own. 
This would be much easier IMO, from the user's POV.

nsarupr: Let me test whether the lowercase can be directly used or not. What's the reasoning behind keeping it in all lower cases?

modules/aws_nodegroup/aws-nodegroup.json
nsarupr: Yeah, the `ami_type` is correct, but not sure whether it should be mentioned in the Documentation. We are keeping this currently support for now, but when we remove this from the usage completely, then we can add this.

RemyDeWolf (approved):

PR #813, opened 2022-04-11
Unhide opta output command

NitinAgg:
# Description

# Safety checklist

YOUR_ANSWER


bigbitbus (approved):
lgtm...are there other hidden=True commands that should be "revealed" in help as well?

PR #820, opened 2022-04-14
Jd/runx module updates

juandiegopalomino:
# Description
Updated the opta-agent helm chart and made it part of the runx module dpeloyment

# Safety checklist

manually

### DIFF
--- a/modules/runx/runx.py
+++ b/modules/runx/runx.py
@@ -32,8 +32,6 @@ else:
 
 class RunxProcessor(ModuleProcessor):
     def __init__(self, module: "Module", layer: "Layer"):
-        if module.data["type"] != "runx":
-            raise Exception(f"The module {module.name} was expected to be of type runx")
         self.user_id = GitConfigParser().get_value("user", "email", "no_user")
         self.device_id = get_mac_address()
         self.os_name = os.name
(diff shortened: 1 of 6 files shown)


modules/runx/runx.py
```diff
@@ -50,10 +48,19 @@ def process(self, module_idx: int) -> None:
...
         current_api_key = self.fetch_secret()
         if current_api_key is None:
-            self.set_secret()
+            current_api_key = self.set_secret()
         else:
             self.fetch_jwt(current_api_key)
 
+        self.module.data["api_key"] = current_api_key
+        k8s_modules = self.layer.get_module_by_type("k8s-cluster", module_idx)
+        if len(k8s_modules) == 0 and self.module.data.get("deploy_k8s_listener") is True:
```
NitinAgg, line 57: Since `deploy_k8s_listener` is true by default. this error could be confusing as a user might not have explicitly set this. Let's just automagically set that to false here instead of an error

NitinAgg (commented):
Minor comment

modules/runx/runx.py
juandiegopalomino: no, because that default does not get applied until later. self.module.data.get("deploy_k8s_listener") returns None if the user has not set it

modules/runx/runx.yaml
```diff
@@ -14,8 +14,15 @@ inputs:
     user_facing: false
     description: Opta Module name
     default: None
+  - name: deploy_k8s_listener
```
RemyDeWolf, line 17: `deploy_opta_agent` sounds more descriptive

RemyDeWolf (approved):

PR #821, opened 2022-04-14
Added cloudfront distribution

bigbitbus:
# Description
Added WAF documentation into cloudfront docs.

# Safety checklist

n/a - its only docs

manually


NitinAgg (approved):
No change to the doc generation script?

PR #825, opened 2022-04-15
Allowing ports to be string so that they can be variable refs

juandiegopalomino:
# Description

# Safety checklist

Manually


bigbitbus (commented):
Can we add validation on the port numbers whent they are strings? Idea for example - https://stackoverflow.com/questions/12968093/regex-to-validate-port-number

juandiegopalomino:
@bigbitbus excellent suggestion

bigbitbus (approved):

PR #826, opened 2022-04-15
Setting kube config and context for helm calls

juandiegopalomino:
# Description
OK, this is what was happening:
We have this nice python class/package to manage our calls to helm. These calls to helm are executed as subprocess whose outputs/error we capture for parsing/error handling. The problem is that these functions are run as part of the k8s service processor, and never specify what kube config and context to use. As nowadays we try to be mindful of the user's kubeconfig and don't overwrite it/set the default unless explicitly asked via opta configure-kubectl, there's no guarantee that the helm commands would even be using the correct kube config or context.
This is worse in CI/CD jobs where because it's being run for the first time, there is no default kubeconfig to speak of. This also explains why the stop-gap of running configure-kubectl before hand works.

How do we fix it?
We know what kubeconfig we're using and what's the name of the ocntext. So use them.

# Safety checklist

manually


legalosLOTR:
What happens if the kube context doesn't exist? Like on a fresh run? Will this change create it as well?

nsarupr:
Makes sense @juandiegopalomino. But why did this became a consistent issue now. Earlier it was working as we wanted it to work.

NitinAgg (approved):

juandiegopalomino:
@legalosLOTR good question-- even in a fresh run the opta-generated kubeconfig is created at the beginning (this is not the default ~/.kube/config, this is a "secondary" which opta keeps track on in ~/.opta/kubeconfigs/), so that is not an issue.
I tested this myself by nuking my default kubeconfig and all opta-generated ones, and it worked as expected, picking up the new one which was generated on that run

legalosLOTR:
got it! lgtm

PR #829, opened 2022-04-15
Bring Your Own Kubernetes: EKS support MVP

RemyDeWolf:
# Description
Allow deploying a k8s service with Opta using the default `~/.kube/config` or the `KUBECONFIG` env var.

This was implemented by adding a Opta registry for Helm. Which means if the user defines a k8s-service and doesn't set a provider or an environment file - we would use their local Kubeconfig to deploy that service with Opta.

Limitations:
- only support k8s-service
- use terraform local backend

We can address these limitations in the future, we have tickets for them in our backlog under the epic _BYOK Extra Features_.

See the[ documentation page](https://github.com/run-x/opta/blob/remy/byok/examples/byok-eks/README.md) for more information - included with this PR.

# Safety checklist

Tested on a EKS cluster created with `k8s-cluster` but NOT with `k8s-base` - instead I ran the terraform files provided with the example and then I ran `opta apply` on this [file](https://github.com/run-x/opta/blob/remy/byok/tests/fixtures/sample_opta_files/byok_service.yaml).

https://www.loom.com/share/ef29c8ca7b26441e8b300f87dd900009?sharedAppSource=personal_library


juandiegopalomino (approved):
lgtm, but maybe add a warning about the limitations?

RemyDeWolf:
@juandiegopalomino I added a warning for the user to save their terraform state files.
For the limitation about which modules are supported, Opta already throws a clear error message to the user such as `ERROR: s3 is not a valid module type`

PR #833, opened 2022-04-15
Use ACM for self signed and imported certs

juandiegopalomino:
# Description
Don't just upload them to ingress nginx

# Safety checklist

Manually

### DIFF
--- a/modules/aws_k8s_base/tf_module/ingress_nginx.tf
+++ b/modules/aws_k8s_base/tf_module/ingress_nginx.tf
@@ -5,14 +5,14 @@ resource "tls_private_key" "self_signed" {
 }
 
 resource "tls_self_signed_cert" "self_signed" {
-  count     = var.expose_self_signed_ssl ? 1 : 0
+  count           = var.expose_self_signed_ssl ? 1 : 0
   key_algorithm   = "RSA"
   private_key_pem = tls_private_key.self_signed[0].private_key_pem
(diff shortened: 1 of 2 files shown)


modules/aws_k8s_base/tf_module/ingress_nginx.tf
```diff
@@ -14,7 +54,7 @@ resource "helm_release" "ingress-nginx" {
         podLabels : {
           "opta-ingress-healthcheck" : "yes"
         }
-        extraArgs : var.private_key != "" || var.expose_self_signed_ssl ? { default-ssl-certificate : "ingress-nginx/secret-tls" } : {}
+        extraArgs : {}
```
RemyDeWolf, line 57: should nginx also use the ACM certificate?
by setting this empty, it will use the _nginx_ self-signed certificate.
This is probably fine - I'm not sure if we would have to rotate it when it expires, of if they take care of it.

juandiegopalomino: Huh, we use the default self-signed cert nginx comes up with all the time for nlb -> nginx connection. Should verify the expration, but that's for another PR

juandiegopalomino: Excellent question

RemyDeWolf (approved):

PR #835, opened 2022-04-16
Adding a max helm history field for k8sservice

juandiegopalomino:
# Description
Cause a problem Will was facing is that he made hundreds of deployments and so the history was obscenely large

# Safety checklist

YOUR_ANSWER


bigbitbus (approved):
why should this be a configurable parameter? Are there too many history entries for a customer?

Why not set default to 0 (infinite)? Its not problemmatic for most people, and gives them all their history...

nsarupr:
Should we keep the default value for `max_history` as `0`, the same as terraform does?

PR #837, opened 2022-04-18
Removed regula from github actions and codebase

bigbitbus:
# Description
Removing regula from GH actions, since we are adopting tfsec.

# Safety checklist

Manually


juandiegopalomino (approved):

nsarupr:
@bigbitbus can you remove this as well.
https://github.com/run-x/opta/blob/feature/remove-regula/.circleci/src/jobs/regula-lint.yaml

PR #841, opened 2022-04-19
Set minimum TLS version for cloudfront

bigbitbus:
# Description

https://aquasecurity.github.io/tfsec/v1.18.0/checks/aws/cloudfront/use-secure-tls-policy/

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/cloudfront_distribution/tf_module/main.tf
+++ b/modules/cloudfront_distribution/tf_module/main.tf
@@ -15,12 +15,12 @@ locals {
 
 resource "aws_cloudfront_distribution" "distribution" {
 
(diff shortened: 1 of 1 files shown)


juandiegopalomino (approved):

bigbitbus:
Actually, we cant set the minimum version, because just read 

 The minimum version of the SSL protocol that you want CloudFront to use for HTTPS connections. Can only be set if cloudfront_default_certificate = false

And we do set this boolean to true under certain cases of user certs.

Here: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/cloudfront_distribution

Instead, will have tfsec ignore this rule.

PR #842, opened 2022-04-19
Allow manual overrides of linked secrets

juandiegopalomino:
# Description

# Safety checklist

manually

### DIFF
--- a/modules/opta-k8s-service-helm/templates/deployment.yaml
+++ b/modules/opta-k8s-service-helm/templates/deployment.yaml
@@ -66,10 +66,10 @@ spec:
             {{ end }}
           envFrom:
           - secretRef:
-              name: manual-secrets
+              name: secret
               optional: true
           - secretRef:
-              name: secrets
+              name: manual-secrets
               optional: true
           {{- if .Values.livenessProbePath }}
           livenessProbe:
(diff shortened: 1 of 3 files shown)


modules/opta-k8s-service-helm/templates/deployment.yaml
```diff
@@ -64,18 +64,13 @@ spec:
...
-                  key: {{ $val.name | quote }}
-                  optional: true
-            {{ end }}
           envFrom:
           - secretRef:
               name: manual-secrets
               optional: true
+          - secretRef:
+              name: secrets
+              optional: true
```
RemyDeWolf, line 73: correct me if I'm wrong here:
`opta secret` would only updates `manual-secrets` and not change `secrets` (which contains the linked secret)

but since `envFrom: secretRef:secrets` is defined last, it will take precedence, so the container would use the old value at runtime.

_envFrom: When a key exists in multiple sources, the value associated with the last source will take precedence._ 
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core

juandiegopalomino: huh, I witnessed the opposite behavior

PR #846, opened 2022-04-20
Adding a series of resource updates to opta

juandiegopalomino:
# Description
1. More default resources to linkerd pods in k8s service
2. Fixed obsolete linkerd HA values yaml to have proper values and set catch-all sidecar resources
3. Added missing linkerd proxy resource annotations to nginx
4. Added resource requests to autoscaler

# Safety checklist

manually

### DIFF
--- a/modules/aws_k8s_base/tf_module/values-ha.yaml
+++ b/modules/aws_k8s_base/tf_module/values-ha.yaml
@@ -4,24 +4,25 @@
 
 enablePodAntiAffinity: true
 
-global:
-  # proxy configuration
-  proxy:
-    resources:
-      cpu:
-        request: 100m
-      memory:
-        limit: 250Mi
-        request: 20Mi
+# We explicitly set the annotations in ingress-nginx and k8s service, but leaving this here for whatever else a user
+# may add.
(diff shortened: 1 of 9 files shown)


rabbitfang (approved):

modules/aws_k8s_base/tf_module/values-ha.yaml
```diff
@@ -30,55 +31,29 @@ publicAPIResources: *controller_resources
 identityResources:
   cpu: *controller_resources_cpu
   memory:
-    limit: 250Mi
+    limit: ""
```
bigbitbus, line 34: This means no limit on controller any longer?

bigbitbus (approved):

modules/aws_k8s_base/tf_module/values-ha.yaml
juandiegopalomino: not needed for this

juandiegopalomino: much rather see it take more than expected memory than fail

PR #849, opened 2022-04-22
Command to remove secrets

juandiegopalomino:
# Description

# Safety checklist

manually

### DIFF
--- a/opta/commands/secret.py
+++ b/opta/commands/secret.py
@@ -7,6 +7,7 @@ from opta.amplitude import amplitude_client
 from opta.commands.apply import local_setup
 from opta.core.generator import gen_all
 from opta.core.kubernetes import (
+    check_if_namespace_exists,
     create_namespace_if_not_exists,
     restart_deployments,
     set_kube_config,
@@ -103,7 +104,7 @@ def view(
             f" already there - update it via secret update."
         )
 
-    logger.info(secrets[secret])
+    print(secrets[secret])
 
 
 @secret.command(name="list")
@@ -189,7 +190,7 @@ def update(
 @env_option
 @input_variable_option
 @local_option
-def remove(
+def delete(
     secret: str,
     env: Optional[str],
     config: str,
@@ -197,11 +198,11 @@ def remove(
     local: Optional[bool],
     var: Dict[str, str],
 ) -> None:
-    """Update a given secret of a k8s service by removing a manual entry
+    """Delete a secret key from a k8s service
 
     Examples:
 
-    opta secret remove -c my-service.yaml "MY_SECRET_1"
+    opta secret delete -c my-service.yaml "MY_SECRET_1"
     """
 
     config = check_opta_file_exists(config)
@@ -212,11 +213,10 @@ def remove(
     gen_all(layer)
 
     set_kube_config(layer)
-    create_namespace_if_not_exists(layer.name)
+    if check_if_namespace_exists(layer.name):
+        remove_manual_secrets(layer.name, secret)
+        __restart_deployments(no_restart, layer.name)
     amplitude_client.send_event(amplitude_client.UPDATE_SECRET_EVENT)
-    remove_manual_secrets(layer.name, secret)
-    __restart_deployments(no_restart, layer.name)
-
     logger.info("Success")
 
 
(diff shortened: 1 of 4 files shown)


opta/commands/secret.py
```diff
@@ -178,7 +179,45 @@ def update(
...
+
+
+@secret.command()
+@click.argument("secret")
+@restart_option
+@config_option
+@env_option
+@input_variable_option
+@local_option
+def remove(
```
RemyDeWolf, line 192: I would suggest delete to be consistent with ` kubectl delete secret`

opta/commands/secret.py
```diff
@@ -178,7 +179,45 @@ def update(
...
+    secret: str,
+    env: Optional[str],
+    config: str,
+    no_restart: bool,
+    local: Optional[bool],
+    var: Dict[str, str],
+) -> None:
+    """Update a given secret of a k8s service by removing a manual entry
```
RemyDeWolf, line 200: Let's stay away from using _update_ since we already have `opta secrets update`
-> `Delete a secret key from a k8s service`

opta/core/kubernetes.py
```diff
@@ -273,6 +273,22 @@ def update_secrets(namespace: str, secret_name: str, new_values: dict) -> None:
     v1.replace_namespaced_secret(secret_name, namespace, current_secret_object)
 
 
+def remove_secrets(namespace: str, secret_name: str, entry_name: str) -> None:
```
RemyDeWolf, line 276: use singular because it's only one secret - or rename `delete_secret_key/entry`

opta/core/secrets.py
```diff
@@ -32,6 +32,15 @@ def update_manual_secrets(layer_name: str, new_values: dict) -> None:
     update_secrets(layer_name, MANUAL_SECRET_NAME, new_values)
 
 
+def remove_manual_secrets(layer_name: str, secret_name: str) -> None:
+    """
+    append the new values to the existing data for this manual secret.
```
RemyDeWolf, line 37: this comment needs to be updated

opta/commands/secret.py
```diff
@@ -102,7 +103,7 @@ def view(
             f" already there - update it via secret update."
         )
 
-    print(secrets[secret])
+    logger.info(secrets[secret])
```
rabbitfang, line 106: I don't think we should use the logger to print the secrets since that would make the line contain extra info (maybe?). I think it would also send the secret data to datadog, which is definitely not something we want.

opta/core/kubernetes.py
```diff
@@ -273,6 +273,22 @@ def update_secrets(namespace: str, secret_name: str, new_values: dict) -> None:
...
+    """
+    remove secret from secret store.
+
+    create the secret if it doesn't exist yet.
+    """
+    load_opta_kube_config()
+    v1 = CoreV1Api()
+    create_secret_if_not_exists(namespace, secret_name)
```
rabbitfang, line 284: Why are we creating a secret resource if we are removing a secret? Shouldn't we return early if it doesn't exist?

opta/core/kubernetes.py
```diff
@@ -273,6 +273,22 @@ def update_secrets(namespace: str, secret_name: str, new_values: dict) -> None:
...
+    current_secret_object: V1Secret = v1.read_namespaced_secret(secret_name, namespace)
+    current_secret_object.data = current_secret_object.data or {}
+    if entry_name in current_secret_object.data:
+        del current_secret_object.data[entry_name]
```
rabbitfang, line 288:
```suggestion
    current_secret_object.data.pop(entry_name, None)
```

Alternative suggestion to remove duplication.

opta/commands/secret.py
```diff
@@ -178,7 +179,45 @@ def update(
...
+
+    config = check_opta_file_exists(config)
+    if local:
+        config = local_setup(config, input_variables=var)
+        env = "localopta"
+    layer = Layer.load_from_yaml(config, env, input_variables=var)
+    gen_all(layer)
+
+    set_kube_config(layer)
+    create_namespace_if_not_exists(layer.name)
```
rabbitfang, line 215: If the namespace doesn't exist, do we really need to create it if the goal is to delete a secret?

rabbitfang (changes requested):

PR #850, opened 2022-04-22
Allow creation of a database by restoring from snapshot

juandiegopalomino:
# Description

# Safety checklist

manually

### DIFF
--- a/modules/aws_postgres/aws-postgres.md
+++ b/modules/aws_postgres/aws-postgres.md
@@ -38,6 +38,9 @@ be set to those of the original database. If you try to link this module to a k8
 secrets will be set to `UNKNOWN_SEE_ORIGINAL_DB`. To overcome this limitation, please use `opta secret update`
 on the your k8s-service yaml to manually populate the secrets in with the original values.
 
+__NOTE__: Do not remove the `restore_from_snapshot` input after creation, even if the snapshot is created as it
+will cause the database to recreate.
+
 ### Adding to Global Database
 
 Opta supports distributed databases via AWS' Aurora Global database. This subject is talked in detail in the 
--- a/modules/aws_postgres/aws-postgres.md
+++ b/modules/aws_postgres/aws-postgres.md
@@ -26,6 +26,29 @@ Storage scaling is automatically managed by AWS Aurora, see the [official docume
 To add replicas to an existing cluser, follow the [official guide](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-replicas-adding.html).
 
 
+### Restoring from Snapshot
+
+Opta supports creating a new database not just from scratch, but from database snapshots you have in your
+desired account + region. To do this, simply set the `restore_from_snapshot` to the ARN of said snapshot you
+wish to restore from.
+
+There is a big caveat: as Opta has no knowledge of the initial db setup from which the snapshot was created, we
+do not know what database name, username, or password the database would be created with-- these will always
+be set to those of the original database. If you try to link this module to a k8s-service, the db name/username/password
+secrets will be set to `UNKNOWN_SEE_ORIGINAL_DB`. To overcome this limitation, please use `opta secret update`
+on the your k8s-service yaml to manually populate the secrets in with the original values.
+
+### Adding to Global Database
+
+Opta supports distributed databases via AWS' Aurora Global database. This subject is talked in detail in the 
+[multi-region section of the docs](/features/multi-region/multi-region-dbs/)
+
+There is a big caveat: as the execution of Opta has no knowledge of the initial db serving as the primary, we
+do not know what database name, username, or password the database would be created with-- these will always
+be set to those of the original database. If you try to link this module to a k8s-service, the db name/username/password
+secrets will be set to `UNKNOWN_SEE_PRIMARY_DB`. To overcome this limitation, please use `opta secret update`
+on the your k8s-service yaml to manually populate the secrets in with the original values.
+
 ### Linking
 
 When linked to a k8s-service, it adds connection credentials to your container's environment variables as:
--- a/modules/aws_postgres/tf_module/outputs.tf
+++ b/modules/aws_postgres/tf_module/outputs.tf
@@ -9,12 +9,12 @@ resource "time_sleep" "wait_for_db" {
 }
 
 output "db_user" {
-  value      = var.existing_global_database_id == null ? aws_rds_cluster.db_cluster[0].master_username : "UNKNOWN_SEE_PRIMARY_DB"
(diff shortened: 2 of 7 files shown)


nsarupr:
Can the similar changes be added to `aws-mysql`? Only difference is the `engine`.
Also, we are keeping the name, user and password as `null`, but the [documentation]()https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster#master_password mentions that we should not add any value to it. So, should we be using the `dynamic block` or is `null` okay.

juandiegopalomino:
@nsarupr excellent questions.
Yes, keepign the password/username/dbname as null for when snapshot restore worked fine
Yes, I'll do an update for mysql, but was thinking of doint it another PR as a bulk update as I'll be adding global db option to it as well.

nsarupr:
> @nsarupr excellent questions. Yes, keepign the password/username/dbname as null for when snapshot restore worked fine Yes, I'll do an update for mysql, but was thinking of doint it another PR as a bulk update as I'll be adding global db option to it as well.

Sounds good @juandiegopalomino

nsarupr (approved):
Looks good to me.

modules/aws_postgres/tf_module/rds.tf
```diff
@@ -21,18 +21,20 @@ resource "aws_rds_cluster" "db_cluster" {
   count                   = var.existing_global_database_id == null ? 1 : 0
   cluster_identifier      = "opta-${var.layer_name}-${var.module_name}-${random_string.db_name_hash.result}"
   db_subnet_group_name    = "opta-${var.env_name}"
-  database_name           = var.database_name
+  database_name           = var.restore_from_snapshot == null ? var.database_name : null
```
rabbitfang, line 24: `database_name` is a valid parameter when restoring from a snapshot (see [`rds:RestoreDBClusterFromSnapshot`](https://docs.aws.amazon.com/AmazonRDS/latest/APIReference/API_RestoreDBClusterFromSnapshot.html), but [the docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/USER_RestoreFromSnapshot.html) don't mention what happens when that is set.

modules/aws_postgres/tf_module/rds.tf
```diff
@@ -21,18 +21,20 @@ resource "aws_rds_cluster" "db_cluster" {
...
   engine                  = "aurora-postgresql"
   engine_version          = var.engine_version
-  master_username         = "postgres"
-  master_password         = random_password.pg_password.result
+  master_username         = var.restore_from_snapshot == null ? "postgres" : null
```
rabbitfang, line 27:
```suggestion
  master_username         = "postgres"
```

`master_username` is ignored when restoring from a snapshot, so keeping the "postgres" value is safe.

modules/aws_postgres/tf_module/rds.tf
```diff
@@ -21,18 +21,20 @@ resource "aws_rds_cluster" "db_cluster" {
...
+  master_password         = var.restore_from_snapshot == null ? random_password.pg_password.result : null
```
rabbitfang, line 28: Setting this value during a snapshot restore will cause the password to be reset after the snapshot is restored.

modules/aws_postgres/tf_module/outputs.tf
```diff
@@ -9,12 +9,12 @@ resource "time_sleep" "wait_for_db" {
...
 
 output "db_user" {
-  value      = var.existing_global_database_id == null ? aws_rds_cluster.db_cluster[0].master_username : "UNKNOWN_SEE_PRIMARY_DB"
+  value      = (var.existing_global_database_id == null && var.restore_from_snapshot == null) ? aws_rds_cluster.db_cluster[0].master_username : (var.restore_from_snapshot == null ? "UNKNOWN_SEE_PRIMARY_DB" : "UNKNOWN_SEE_ORIGINAL_DB")
   depends_on = [time_sleep.wait_for_db]
 }
 
 output "db_password" {
-  value      = var.existing_global_database_id == null ? aws_rds_cluster.db_cluster[0].master_password : "UNKNOWN_SEE_PRIMARY_DB"
+  value      = (var.existing_global_database_id == null && var.restore_from_snapshot == null) ? aws_rds_cluster.db_cluster[0].master_password : (var.restore_from_snapshot == null ? "UNKNOWN_SEE_PRIMARY_DB" : "UNKNOWN_SEE_ORIGINAL_DB")
```
rabbitfang, line 17: Is this going to cause issues if the DB password is referenced by other modules? Ditto for the DB user and name, but the password is probably the most significant.

rabbitfang (commented):
I think we should support unsetting the `restore_from_snapshot` variable. I can anticipate users not wanting to keep a snapshot ID around in their opta config for the life of the DB. Once the DB is restored, they should be able to safely drop that variable. If we don't support that, at the very least we should make it clear in the documentation that the variable cannot be unset.

My comments below talk about some of the details of supporting that use case.

modules/aws_postgres/tf_module/rds.tf
juandiegopalomino: When it's not set then it inherits the original db's name, which I think is the behavior we want and I don't see a reason why we ought to support some renaming. It jus feels wrong and worse case scenario it can be done so via sql commands

modules/aws_postgres/tf_module/rds.tf
juandiegopalomino: source?

modules/aws_postgres/tf_module/rds.tf
juandiegopalomino: source?

modules/aws_postgres/tf_module/outputs.tf
juandiegopalomino: excellent question
the intent was that if a db was being restored then they would use manual secrets to get the username/db name/password from the old deployment

modules/aws_postgres/tf_module/rds.tf
rabbitfang: [AWS provider source code](https://github.com/hashicorp/terraform-provider-aws/blob/8b5a71bcdcaa04e6d5c5c41243fd981edbdd79a5/internal/service/rds/cluster.go#L542). Basically, `master_username` is not read when restoring from snapshot, but it *is* marked as `ForceNew`, so if it changes (such as if the `restore_from_snapshot` variable is unset), it will force a recreation of the aurora cluster.

modules/aws_postgres/tf_module/rds.tf
rabbitfang: Same source as `master_username`

modules/aws_postgres/aws-postgres.md
```diff
@@ -26,6 +26,29 @@ Storage scaling is automatically managed by AWS Aurora, see the [official docume
 To add replicas to an existing cluser, follow the [official guide](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/aurora-replicas-adding.html).
 
 
+### Restoring from Snapshot
```
rabbitfang, line 29: I would add a note that opta doesn't support removing this value after the snapshot has been restored.

PR #851, opened 2022-04-22
Gitlab opta installation

nsarupr:
# Description
Install Gitlab using Opta.

# Safety checklist

Tested manually.
gitlab-login
gitlab-dashboard


examples/gitlab/README.md
```diff
@@ -0,0 +1,54 @@
...
+* Go to your domain registrar (link namecheap, godaddy, etc.) to point the domain to these nameservers.
+* Update `delegated` field to `true` in the `gitlab-aws.yaml` file
+* Run `opta apply -c gitlab-aws.yaml` again to generate the TLS certificate
+
+Your domain should now be pointing to the Ghost deployment with secure TLS
+
+* The admin dashboard would be available at https://gitlab.{your-domain}/
+* To find the password for the root user:
+```shell
+(opta) nileshsarupriya@Nileshs-MacBook-Pro ~/D/r/opta (main)> echo $(kubectl get secrets -n gitlab gitlab-gitlab-gitlab-initial-root-password -o json | jq '.data.password')
```
juandiegopalomino, line 42: don't show your shell add-ons (e.g. (opta) nileshsarupriya@Nileshs-MacBook-Pro ~/D/r/opta (main)>), instead just put "$"

examples/gitlab/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+Your domain should now be pointing to the Ghost deployment with secure TLS
+
+* The admin dashboard would be available at https://gitlab.{your-domain}/
+* To find the password for the root user:
+```shell
+$> echo $(kubectl get secrets -n gitlab gitlab-gitlab-gitlab-initial-root-password -o json | jq '.data.password')
+"OTdjNklub0wzOFlYVWRRSENhRWJhMzRZeHRkNEFmV0NoWjVZcmw0VDlMTXp5eDlBYldrZXBvenJERWxwM2pxcQ=="
+$> echo "OTdjNklub0wzOFlYVWRRSENhRWJhMzRZeHRkNEFmV0NoWjVZcmw0VDlMTXp5eDlBYldrZXBvenJERWxwM2pxcQ==" | base64 -D
+97c6InoL38YXUdQHCaEba34Yxtd4AfWChZ5Yrl4T9LMzyx9AbWkepozrDElp3jqq
```
rabbitfang, line 45:
```suggestion
$ kubectl get secrets -n gitlab gitlab-gitlab-gitlab-initial-root-password -o json | jq -r '.data.password' | base64 -D
example_password
```

- `echo $( ... )` is superfluous
- Using "real" secret values, even in examples is generally not a good idea because sometimes people copy those example values and use them for real, and automated scanners can pick them up (and then "report them" to us in an issue)
- Its better to just keep piping the data through instead of making the user copy/paste

examples/gitlab/README.md
```diff
@@ -0,0 +1,54 @@
...
+```
+
+![Alt text](gitlab-login.png?raw=true "Login Screen")
+
+![Alt text](gitlab-dashboard.png?raw=true "What it should look like")
```
rabbitfang, line 50:
```suggestion
![GitLab Login Screen](gitlab-login.png?raw=true "GitLab Login Screen")

![What it should look like](gitlab-dashboard.png?raw=true "What it should look like")
```

Alt text should be set even if its the same as the image title

examples/gitlab/gitlab-aws.yaml
```diff
@@ -0,0 +1,60 @@
...
+    chart_version: "5.8.2"
+    values:
+      postgresql:
+        master:
+          podAnnotations:
+            linkerd.io/inject: enabled
+            config.linkerd.io/proxy-cpu-limit: "0.1"
+            config.linkerd.io/proxy-cpu-request: "0.05"
+            config.linkerd.io/proxy-memory-limit: "20Mi"
+            config.linkerd.io/proxy-memory-request: "10Mi"
```
rabbitfang, line 33:
```suggestion
```

These values are now set by default, so the user doesn't need to set these explicitly.

examples/gitlab/gitlab-aws.yaml
```diff
@@ -0,0 +1,60 @@
...
+      gitlab-runner:
+        podAnnotations:
+          linkerd.io/inject: enabled
+          config.linkerd.io/proxy-cpu-limit: "0.1"
+          config.linkerd.io/proxy-cpu-request: "0.05"
+          config.linkerd.io/proxy-memory-limit: "20Mi"
+          config.linkerd.io/proxy-memory-request: "10Mi"
```
rabbitfang, line 40:
```suggestion
```

Ditto

examples/gitlab/gitlab-aws.yaml
```diff
@@ -0,0 +1,60 @@
...
+          class: nginx
+        # psql:
+        #   host: "db_host"
+        #   port: 5432
+        #   username: "db_user"
+        #   database: "db_name"
+        #   password:
+        #     useSecret: true
+        #     secret: gitlab-postgres
+        #     key: psql-password
```
rabbitfang, line 58:
```suggestion
```

Shouldn't leave these commented out sections in the example.

examples/gitlab/gitlab-aws.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - type: base
+  - type: dns
+    domain: <<domain::string::The dns domain for your airflow deployment>>
+    delegated: false
+  - type: k8s-cluster
+  - type: k8s-base
+  # - type: postgres
+  #   name: gitlabpg
+  # - type: redis
+  #   name: gitlabredis
```
rabbitfang, line 17:
```suggestion
```

If the commented out section isn't needed as part of the example, it should be removed.

examples/gitlab/gitlab-aws.yaml
```diff
@@ -0,0 +1,60 @@
...
+        #   host: "db_host"
+        #   port: 5432
+        #   username: "db_user"
+        #   database: "db_name"
+        #   password:
+        #     useSecret: true
+        #     secret: gitlab-postgres
+        #     key: psql-password
+      certmanager-issuer:
+        email: nilesh@runx.dev
```
rabbitfang, line 60: A real email shouldn't be used here. Should use something like `foo@example.com`.

rabbitfang (changes requested):

examples/gitlab/gitlab-aws.yaml
nsarupr: These values are needed to be added, as pod annotations for linkerd works only for it's own pods, nowhere else.

examples/gitlab/gitlab-aws.yaml
nsarupr: I kept these as commented to let the users know that RDS and Redis can be used directly as such

examples/gitlab/README.md
nsarupr: I'll change this

examples/gitlab/README.md
nsarupr: passing values to pipe was essentially failing, and I didn't want this to become a bad user experience.
As for the actual values, I'll replace them with dummy.

rabbitfang: What was "essentially failing"? I added the `-r` flag to `jq` (output raw values) which drops the quote marks from the output. Other than that, I can't think of what would cause failures.

examples/gitlab/gitlab-aws.yaml
rabbitfang: There was a [recent change](https://github.com/run-x/opta/pull/846/files#diff-64bb93a8e0db0a2018722bfe687556f70adb7564249961b0a16c3c93103b94d8L7-R16) that should fix the issue where linkerd sidecar containers don't receive a default set of resource requirements. Do those changes not work as expected?

rabbitfang (changes requested):
Can you change your screenshots to hide the bookmark tab and your installed extensions? You can crop the bookmark tab out and just do a solid fill rectangle over the extension icons (matching the rest of the menu bar background)

rabbitfang (approved):

PR #852, opened 2022-04-22
Updating to non-deprecated terraform field for source json

juandiegopalomino:
# Description

# Safety checklist

YOUR_ANSWER

### DIFF
--- a/modules/aws_s3/tf_module/bucket.tf
+++ b/modules/aws_s3/tf_module/bucket.tf
@@ -96,7 +96,7 @@ resource "aws_cloudfront_origin_access_identity" "read" {
 
 
 data "aws_iam_policy_document" "s3_policy" {
-  source_json = var.bucket_policy == null ? "" : jsonencode(var.bucket_policy)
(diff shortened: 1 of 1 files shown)


modules/aws_s3/tf_module/bucket.tf
```diff
@@ -96,7 +96,7 @@ resource "aws_cloudfront_origin_access_identity" "read" {
 
 
 data "aws_iam_policy_document" "s3_policy" {
-  source_json = var.bucket_policy == null ? "" : jsonencode(var.bucket_policy)
+  source_policy_documents = var.bucket_policy == null ? [] : jsonencode(var.bucket_policy)
```
rabbitfang, line 99:
```suggestion
  source_policy_documents = var.bucket_policy == null ? [] : [jsonencode(var.bucket_policy)]
```

`source_policy_documents` is supposed to be a list of documents, so I think we should wrap the value returned from `jsonencode`

rabbitfang (approved):

PR #858, opened 2022-04-29
Add AWS VPN Module

juandiegopalomino:
# Description
Adds AWS VPN module for Opta. See md for full description

# Safety checklist

manually

### DIFF
--- a/modules/aws_vpn/aws-vpn.md
+++ b/modules/aws_vpn/aws-vpn.md
@@ -4,7 +4,7 @@ linkTitle: "aws-vpn"
 date: 2022-04-28
 draft: false
 weight: 1
-description: Creates an AWS VPN endpoint for folks to access your private Opta network
+description: Creates an AWS VPN endpoint for folks to access their private Opta network
 ---
 
 This module sets up an [AWS Client VPN Endpoint](https://aws.amazon.com/vpn/client-vpn/) for you. For those new to
@@ -36,17 +36,16 @@ For upcoming work, it's best to place it just after the base module instead of a
 Run `opta apply` and you should be good to go!
 
 ## Setting Up the Client
-Short story:
 AWS VPN uses the OpenVPN client and the ovpn file format to set up the configuration for the connection profile. Opta
-makes a default profile for you and store is in an SSM parameter which you can fetch it from (refer to the 
+makes a default profile for you and stores it in an SSM parameter which you can fetch it from (refer to the 
 `ovpn_profile_parameter_arn` output). Simply fetch the value from this parameter, write it into a .ovpn file locally
-and tell your OpenVPN client to import a new profile from said file. And that's it!
+and tell your OpenVPN client to import a new profile from said file.
 
-Long Story:
 Our VPN uses mTLS to handle the network encryption and validation of both the server and each client connecting to it.
 Opta creates a brand new Certificate Authority to provision the certs for both the servers and clients. As in larger
 companies it's an important security consideration to revoke VPN access on a per-person use case, it's highly 
-recommended to generate a new client key/cert for each user. This can be done in the following steps:
+recommended to generate a new client key/cert for each user. This can be done in the following steps (steps 2-5
+will be repeated for each new user):
 
 1. Download the certificate and key of the CA created for your VPN. Again, we store them as SSM parameters whose 
    arns are the outputs `vpn_ca_cert_parameter_arn` and `vpn_ca_key_parameter_arn`. Name them `rootCA.crt` and
@@ -78,6 +77,6 @@ with the new key and cert just created. Create a new OpenVPN profile from it and
 * You can monitor current connections from thr AWS console by going to the VPN endpoint resource.
 
 ## Caveats
-* Note that the default client certificate and the one displayed above is set to expire after 6 months. This short
+* Note that the default client certificate and the one displayed above is set to expire after 2 years. This short
   lifetime is intentional due to the security risk of having a VPN profile be valid for too long.
 * You will need to use certificate revocation lists to revoke individual certificates.
--- a/modules/aws_vpn/aws-vpn.md
+++ b/modules/aws_vpn/aws-vpn.md
@@ -1,0 +1,41 @@
+---
+title: "aws-vpn"
+linkTitle: "aws-vpn"
+date: 2022-04-28
+draft: false
+weight: 1
+description: Creates an AWS VPN endpoint for folks to access your private Opta network
+---
+
+This module sets up an [AWS Client VPN Endpoint](https://aws.amazon.com/vpn/client-vpn/) for you. For those new to
+this service, AWS Client VPN is AWS' implementation of a VPN which natively integrates to their networking solutions.
+It follows the OpenVPN specification and can readily be connected to with the [OpenVPN Client](https://openvpn.net/vpn-client/)
+(which we recommend).
+
+As with other VPN solutions, this one is used to gain secure access to a private network, in this case the private VPC
+subnets Opta provisioned for your environment. This means that when connected you should have the network access needed
+to locally connect to your databases, caches, pretty much all resources in your private network that are accepting 
+traffic.
+
+## Adding VPN to your Environment
+For most cases, it's actually just adding a single line for the VPN module like so:
+```yaml
+name: testing-vpn
+org_name: runx
+providers:
+  aws:
+    region: us-east-1
+    account_id: XXXXXXXXXX
+modules:
+  - type: base
+  - type: aws-vpn
+  - type: k8s-cluster
+  - type: k8s-base
+```
+For upcoming work, it's best to place it just after the base module instead of at the end.
+Run `opta apply` and you should be good to go!
+
+## Setting Up the Client
+
+
+## Caveats
--- a/modules/aws_vpn/aws-vpn.md
+++ b/modules/aws_vpn/aws-vpn.md
@@ -36,6 +36,48 @@ For upcoming work, it's best to place it just after the base module instead of a
 Run `opta apply` and you should be good to go!
 
 ## Setting Up the Client
+Short story:
+AWS VPN uses the OpenVPN client and the ovpn file format to set up the configuration for the connection profile. Opta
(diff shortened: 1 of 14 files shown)


modules/aws_vpn/aws-vpn.md
```diff
@@ -0,0 +1,83 @@
+---
+title: "aws-vpn"
+linkTitle: "aws-vpn"
+date: 2022-04-28
+draft: false
+weight: 1
+description: Creates an AWS VPN endpoint for folks to access your private Opta network
```
legalosLOTR, line 7: **their private Opta network

modules/aws_vpn/aws-vpn.md
```diff
@@ -0,0 +1,83 @@
...
+```
+with the new key and cert just created. Create a new OpenVPN profile from it and you're done!
+
+## Extra Info
+* You will actually get logs of the VPN activity sent to AWS Cloudwatch under a log group made just for the VPN logs
+  ("opta-ENV_NAME-vpn-RANDOM_STRING").
+* You can monitor current connections from thr AWS console by going to the VPN endpoint resource.
+
+## Caveats
+* Note that the default client certificate and the one displayed above is set to expire after 6 months. This short
```
legalosLOTR, line 81: Let's make the default 2 years at least. 6mo is too short. Also what would be the process to renew? Can we document that or provide a link?

modules/aws_vpn/aws-vpn.md
```diff
@@ -0,0 +1,83 @@
...
+  - type: k8s-cluster
+  - type: k8s-base
+```
+For upcoming work, it's best to place it just after the base module instead of at the end.
+Run `opta apply` and you should be good to go!
+
+## Setting Up the Client
+Short story:
+AWS VPN uses the OpenVPN client and the ovpn file format to set up the configuration for the connection profile. Opta
+makes a default profile for you and store is in an SSM parameter which you can fetch it from (refer to the
```
legalosLOTR, line 41: **stores it in

modules/aws_vpn/aws-vpn.md
```diff
@@ -0,0 +1,83 @@
...
+Run `opta apply` and you should be good to go!
+
+## Setting Up the Client
+Short story:
+AWS VPN uses the OpenVPN client and the ovpn file format to set up the configuration for the connection profile. Opta
+makes a default profile for you and store is in an SSM parameter which you can fetch it from (refer to the 
+`ovpn_profile_parameter_arn` output). Simply fetch the value from this parameter, write it into a .ovpn file locally
+and tell your OpenVPN client to import a new profile from said file. And that's it!
+
+Long Story:
```
legalosLOTR, line 45: I don't think you should split these as short/long. Both seem kinda necessary to me. Maybe just put both as separate paragraphs?

modules/aws_vpn/aws-vpn.md
```diff
@@ -0,0 +1,83 @@
...
+Our VPN uses mTLS to handle the network encryption and validation of both the server and each client connecting to it.
+Opta creates a brand new Certificate Authority to provision the certs for both the servers and clients. As in larger
+companies it's an important security consideration to revoke VPN access on a per-person use case, it's highly 
+recommended to generate a new client key/cert for each user. This can be done in the following steps:
+
+1. Download the certificate and key of the CA created for your VPN. Again, we store them as SSM parameters whose
```
legalosLOTR, line 51: Can you clarify that steps 2-4 need to be followed for each user?
Also, what are the steps to revoke a user's certificates?

modules/aws_vpn/tf_module/cert.tf
```diff
@@ -0,0 +1,137 @@
...
+
+resource "aws_ssm_parameter" "ca_key" {
+  name        = "/opta/${var.env_name}/vpn-ca-key-pem"
+  description = "The pem of the key of the VPN CA certificate"
+  type        = "SecureString"
+  tier        = "Advanced"
+  value       = tls_private_key.ca_key.private_key_pem
+}
+
+resource "tls_private_key" "server_issuer_key" {
```
rabbitfang, line 37:
```suggestion
resource "tls_private_key" "server_key" {
```

modules/aws_vpn/tf_module/cert.tf
```diff
@@ -0,0 +1,137 @@
...
+  algorithm = "RSA"
+  rsa_bits  = "2048"
+}
+
+resource "tls_cert_request" "server_issuer_req" {
+  key_algorithm   = tls_private_key.server_issuer_key.algorithm
+  private_key_pem = tls_private_key.server_issuer_key.private_key_pem
```
rabbitfang, line 44:
```suggestion
resource "tls_cert_request" "server_req" {
  key_algorithm   = tls_private_key.server_key.algorithm
  private_key_pem = tls_private_key.server_key.private_key_pem
```

modules/aws_vpn/tf_module/cert.tf
```diff
@@ -0,0 +1,137 @@
...
+  dns_names       = ["server"]
+
+  subject {
+    common_name = "server"
+  }
+}
+
+resource "tls_locally_signed_cert" "server_issuer_cert" {
+  cert_request_pem      = tls_cert_request.server_issuer_req.cert_request_pem
```
rabbitfang, line 53:
```suggestion
resource "tls_locally_signed_cert" "server_cert" {
  cert_request_pem      = tls_cert_request.server_req.cert_request_pem
```

modules/aws_vpn/tf_module/cert.tf
```diff
@@ -0,0 +1,137 @@
...
+resource "aws_ssm_parameter" "ovpn_profile" {
+  name        = "/opta/${var.env_name}/ovpn"
+  description = "The vpn profile for the Opta-provisioned AWS VPN endpoint for this environment"
+  type        = "SecureString"
+  tier        = "Advanced"
+  value       = <<EOT
+client
+dev tun
+proto udp
+remote opta.${trimprefix(aws_ec2_client_vpn_endpoint.vpn.dns_name, "*.")} 443
```
rabbitfang, line 117:
```suggestion
remote ${trimprefix(aws_ec2_client_vpn_endpoint.vpn.dns_name, "*.")} 443
```

`remote-random-hostname` causes the remote hostname to be prefixed with a random subdomain. Including `opta.` might cause DNS issues

modules/aws_vpn/aws_vpn.py
```diff
@@ -0,0 +1,47 @@
...
+        self.module.data["vpc_id"] = self.module.data.get(
+            "vpc_id", f"${{{{{base_module_source}.vpc_id}}}}"
+        )
+        self.module.data["public_subnets_ids"] = self.module.data.get(
+            "public_subnets_ids", f"${{{{{base_module_source}.public_subnets_ids}}}}"
+        )
+        self.module.data["kms_account_key_arn"] = self.module.data.get(
+            "kms_account_key_arn", f"${{{{{base_module_source}.kms_account_key_arn}}}}"
+        )
+        super(AwsVPNProcessor, self).process(module_idx)
```
rabbitfang, line 47:
```suggestion
        super().process(module_idx)
```

The arguments to `super()` are optional in Python 3, and should usually be omitted.

modules/aws_vpn/tf_module/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  egress {
+    description = "alloutbound"
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    # To be fixed - for now user can create an SG manually to override.
+    #tfsec:ignore:aws-vpc-no-public-egress-sgr
+    cidr_blocks = ["0.0.0.0/0"]
```
rabbitfang, line 25: We could change this to default to the VPC CIDR blocks and allow the user to override to `0.0.0.0/0` if they want to use the VPN for internet access. That would get rid of the tfsec warning.

opta/layer.py
```diff
@@ -463,30 +464,19 @@ def gen_tf(
...
-                        output_prefix=output_prefix,
-                        existing_defaults=existing_defaults,
-                    ),
-                    ret,
-                )
-            except Exception as e:
-                # I dunno why we were swallowing this
-                raise e
-            if module.desc.get("halt"):
-                previous_module_reference = [f"module.{module.name}"]
```
rabbitfang, line 489: Can you explain why these lines were removed?

modules/aws_vpn/aws-vpn.md
rabbitfang, line 7:
```suggestion
description: Creates an AWS Client VPN for accessing private VPC resources
```

modules/aws_vpn/tf_module/cert.tf
```diff
@@ -0,0 +1,137 @@
...
+resource "tls_cert_request" "client_issuer_req" {
+  key_algorithm   = tls_private_key.client_issuer_key.algorithm
+  private_key_pem = tls_private_key.client_issuer_key.private_key_pem
+
+  subject {
+    common_name = "client1.domain.tld"
+  }
+}
+
+resource "tls_locally_signed_cert" "client_issuer_cert" {
```
rabbitfang, line 89: This cert/key is being used as the client cert, so we shouldn't be calling it an "issuer" nor should it have the CA flag set (`is_ca_certificate`)

modules/aws_vpn/tf_module/cert.tf
juandiegopalomino: no, that's precisely on purpose, trust me

opta/layer.py
juandiegopalomino: Yes, we've gone a long way in cleaning up implied dependencies and now we don't need them (they often cause cyclical paths during development and unneeded re-reads of data)

modules/aws_vpn/tf_module/main.tf
juandiegopalomino: we're currently hard coding split tunneling so I'm just setting this to the vpc cidr

rabbitfang (approved):

PR #861, opened 2022-04-29
Make sure linkerd and nginx are installed for BYOK clusters

rabbitfang:
# Description
- Checks that nginx ingress and linkerd are installed when running for BYOK clusters (when using the helm cloud client)
- Checking for nginx ingress validates that the `nginx` IngressClass exists
- Checking for linkerd makes sure that a namespace labeled with `linkerd.io/is-control-plane=true` exists.

# Safety checklist

Manually verified that the check fails when either linkerd namespace not labeled correctly or nginx ingressclass removed (or both).


juandiegopalomino (approved):
lgtm

PR #863, opened 2022-04-30
Add support for secrets for helm charts

juandiegopalomino:
# Description
Treat them exactly as you would for k8s-service. Even works for when a k8s-service and helm-chart are in the same yaml, or multiple helm charts in the same yaml.

# Safety checklist

manually

### DIFF
--- a/opta/commands/secret.py
+++ b/opta/commands/secret.py
@@ -1,25 +1,21 @@
-from typing import Dict, Optional
+from typing import Dict, Optional, Tuple
 
 import click
 from click_didyoumean import DYMGroup
 
 from opta.amplitude import amplitude_client
 from opta.commands.apply import local_setup
-from opta.core.generator import gen_all
 from opta.core.kubernetes import (
     check_if_namespace_exists,
     create_namespace_if_not_exists,
+    delete_secret_key,
     restart_deployments,
     set_kube_config,
+    update_secrets,
 )
-from opta.core.secrets import (
-    bulk_update_manual_secrets,
-    delete_manual_secret,
-    get_secrets,
-    update_manual_secrets,
-)
+from opta.core.secrets import bulk_update_manual_secrets, get_secrets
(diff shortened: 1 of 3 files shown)


opta/commands/secret.py
```diff
@@ -39,9 +35,42 @@
...
+            type=click.Choice([x.name for x in total_modules]),
+        )
+    if module_name is None:
+        module: Module = total_modules[0]
+    else:
+        module = layer.get_module(module_name)  # type: ignore
+        if module is None:
+            raise UserErrors(
+                f"Could not find helm/k8s-service module with name {module_name}"
+            )
```
rabbitfang, line 61:
```suggestion
        try:
            module = next(module for module in total_modules if module.name == module_name)
        except StopIteration:
            raise UserErrors(
                f"Could not find helm/k8s-service module with name {module_name}"
            ) from None
```

Disabling type checking should be avoided wherever possible.

opta/commands/secret.py
```diff
@@ -39,9 +35,42 @@
...
+    "-m", "--module", default=None, help="The module to get the secret for"
+)
+
+
+def get_secret_name_and_namespace(
+    layer: Layer, module_name: Optional[str]
+) -> Tuple[str, str]:
+    k8s_services = layer.get_module_by_type("k8s-service")
+    helm_charts = layer.get_module_by_type("helm-chart")
+    total_modules = k8s_services + helm_charts
```
rabbitfang, line 48:
```suggestion
    total_modules = k8s_services + helm_charts
    
    if not total_modules:
        raise UserErrors(f"No helm/k8s-service modules were configured")
```

Should make sure to handle the case where there are no k8s-service or helm-chart modules (unless I am missing where that check is defined elsewhere).

rabbitfang (changes requested):

rabbitfang (approved):

PR #871, opened 2022-05-09
Example to demonstrate multi-region capability

nsarupr:
# Description
Demonstrate the Multi-region capability using Opta.

# Safety checklist

Tested by creating Primary and Secondary environments and respective read/write and read databases


NitinAgg (approved):
This is great. Let's merge this and then I will try this myself sometime this/next week and we can put this on hacker news

PR #875, opened 2022-05-10
Forbid Micro and Nano instances in aws eks

nsarupr:
# Description
Forbid users from using micro and nano instances in AWS EKS.

# Safety checklist

Tested and validated manually.

```yaml
name: testenv
org_name: runx
providers:
  aws:
    region: us-east-1
    account_id: 000000000000
modules:
  - type: base
  - type: k8s-cluster
    node_instance_type: "t2.nano"
  - type: k8s-base
```

```bash
$> opta apply -c aws-env.yml
Opta file validation failed with errors:
  : modules.1: node_instance_type: t2.nano is not a regex match.
ERROR: aws-env.yml is not a valid Opta file.
If you need more help please reach out to the contributors in our slack channel at: https://slack.opta.dev
```


juandiegopalomino (commented):
also on the nodegroup

nsarupr:
> also on the nodegroup

I thought there we should let the users have an option to have a lesser instance type

PR #878, opened 2022-05-13
Adding K8s remote backend to BYOK

juandiegopalomino:
# Description
Now byo k8s cluster will store its states using k8s secrets via the official k8s backend

# Safety checklist

manually

### DIFF
--- a/config/registry/helm/index.yaml
+++ b/config/registry/helm/index.yaml
@@ -3,9 +3,9 @@ required_providers:
     source: "hashicorp/helm"
     version: "2.4.1"
 backend:
-  local:
-    # this is consistent with terraform generator
-    path: "./tfstate/{layer_name}.tfstate"
+  kubernetes:
+    secret_suffix: "{state_storage}"
+    config_path: "{kubeconfig}"
 validator:
   name: str()
   org_name: regex('^[a-z0-9-]{,15}$', name="Valid identifier, regex='[a-z0-9-]{,15}'")
@@ -22,3 +22,5 @@ output_providers:
   helm:
     kubernetes:
       config_path: "{kubeconfig}"
+  kubernetes:
+    config_path: "{kubeconfig}"
--- a/config/registry/helm/index.yaml
+++ b/config/registry/helm/index.yaml
@@ -2,6 +2,9 @@ required_providers:
   helm:
     source: "hashicorp/helm"
     version: "2.4.1"
+  kubernetes:
+    source: "hashicorp/kubernetes"
+    version: "2.11.0"
 backend:
   kubernetes:
     secret_suffix: "{state_storage}"
--- a/opta/core/helm_cloud_client.py
+++ b/opta/core/helm_cloud_client.py
@@ -125,7 +125,6 @@ class HelmCloudClient(CloudClient):
         with open(kube_config_file_name, "w") as f:
             yaml.dump(cluster_config, f)
         constants.GENERATED_KUBE_CONFIG = kube_config_file_name
-        return
 
     def cluster_exist(self) -> bool:
         # "kubectl version" returns an error code if it can't connect to a cluster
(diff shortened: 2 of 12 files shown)


config/registry/helm/index.yaml
```diff
@@ -2,10 +2,13 @@ required_providers:
...
+  kubernetes:
+    source: "hashicorp/kubernetes"
+    version: "2.11.0"
 backend:
-  local:
-    # this is consistent with terraform generator
-    path: "./tfstate/{layer_name}.tfstate"
+  kubernetes:
+    secret_suffix: "{state_storage}"
+    config_path: "{kubeconfig}"
```
rabbitfang, line 11: How is this change going to affect existing usage of the helm cloud provider?

rabbitfang (changes requested):

opta/core/helm_cloud_client.py
```diff
@@ -16,39 +27,114 @@ def __init__(self, layer: "Layer"):
...
+
         return None
 
     def delete_opta_config(self) -> None:
-        return None
+        set_kube_config(self.layer)
+        load_opta_kube_config()
+        v1 = CoreV1Api()
+        secret_name = f"opta-config-{self.layer.state_storage()}"
+        if check_if_secret_exists("default", secret_name):
```
RemyDeWolf, line 64: may be we should allow configuring this namespace to something lese than "default" (which would be the default value if not set)

opta/core/helm_cloud_client.py
```diff
@@ -16,39 +27,114 @@ def __init__(self, layer: "Layer"):
...
+            v1.delete_namespaced_secret(secret_name, "default")
 
     def delete_remote_state(self) -> None:
-        return None
+        v1 = CoreV1Api()
+        secret_name = f"tfstate-default-{self.layer.state_storage()}"
```
RemyDeWolf, line 69: no need to have `default` here since it's a namespace scoped resource

RemyDeWolf (commented):
1. There is a 1MB limit for k8s secret, is that enough for us? (how much do we use for a typical service?

2. Don't forget to update the documentation, the part with:
_Generate the terraform state files in tfstate - please commit these files after each apply._
https://github.com/run-x/opta/tree/main/examples/byok-eks#deploy-a-service-to-kubernetes-with-opta__

opta/core/helm_cloud_client.py
juandiegopalomino: Not for now I think-- default has the benefit of always being present/available

opta/core/helm_cloud_client.py
juandiegopalomino: That's how terraform names the secret-- we got no choice :(

opta/core/helm_cloud_client.py
RemyDeWolf: ok fair enough,  helm also uses to this namespace so we are consistent with that

juandiegopalomino:
@RemyDeWolf 
There is a 1MB limit for k8s secret, is that enough for us? (how much do we use for a typical service?
Yes, absolutely

Don't forget to update the documentation,
Done

RemyDeWolf (approved):

opta/core/helm_cloud_client.py
```diff
@@ -80,38 +81,60 @@ def get_all_remote_configs(self) -> Dict[str, Dict[str, "StructuredConfig"]]:
...
             raise UserErrors(
                 "To use K8s-native provider, your must have your kubeconfig set"
             )
         logger.debug("Loading kube config file")
         try:
-            default_kube_config = yaml.load(open(default_kube_config_filename))
-        except Exception:
+            with open(default_kube_config_filename) as f:
+                default_kube_config = yaml.load(f)
+        except YAMLError:
```
rabbitfang, line 95:
```suggestion
        except yaml.YAMLError:
```

(as opposed to `from ruamel.yaml import YAMLError` above). The `yaml` util library should be used instead of importing ruamel yaml directly, to discourage unsafe usage. That library reexports YAMLError.

rabbitfang (approved):
Minor suggestion, otherwise LGTM

PR #888, opened 2022-05-18
Fixing Cloudfront

juandiegopalomino:
# Description
* Stop adding ssl to port 80 when using self signed certs
* Stop ssl redirect when using self signed certs
* Cloudfront ssl origin does not work if origin's ssl certs are self signed, so updating that

# Safety checklist

Manually

### DIFF
--- a/modules/cloudfront_distribution/cloudfront-distribution.yaml
+++ b/modules/cloudfront_distribution/cloudfront-distribution.yaml
@@ -94,6 +94,11 @@ inputs:
     validator: str(required=False)
     description: ID of Route53 hosted zone to add a record for. By default uses the one created by the DNS module if the module is found.
     default: ""
+  - name: forward_https
+    user_facing: true
+    validator: bool(required=False)
+    description: Should cloudformation forward to https port 443 instead of http port 80?
+    default: false
 extra_validators: { }
 outputs:
   - name: cloudfront_domain
--- a/modules/cloudfront_distribution/tf_module/main.tf
+++ b/modules/cloudfront_distribution/tf_module/main.tf
@@ -21,7 +21,7 @@ resource "aws_cloudfront_distribution" "distribution" {
   enabled         = true
   is_ipv6_enabled = true
   price_class     = var.price_class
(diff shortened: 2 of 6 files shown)


modules/cloudfront_distribution/tf_module/main.tf
```diff
@@ -21,7 +21,7 @@ resource "aws_cloudfront_distribution" "distribution" {
   enabled         = true
   is_ipv6_enabled = true
   price_class     = var.price_class
-  aliases         = var.acm_cert_arn == "" ? [] : var.domains
+  aliases         = var.acm_cert_arn == "" ? [] : concat(var.domains, formatlist("*.%s", var.domains))
```
rabbitfang, line 24: The domains here are required to all be listed on the certificate served by the distribution, so we might not want to add wildcards for every domain by default (see https://docs.amazonaws.cn/en_us/AmazonCloudFront/latest/DeveloperGuide/CNAMEs.html#alternate-domain-names-wildcard). Could we make this optional or just have the user specify which domains should have wildcards atttached themselves?

modules/cloudfront_distribution/cloudfront-distribution.yaml
```diff
@@ -94,6 +94,11 @@ inputs:
     validator: str(required=False)
     description: ID of Route53 hosted zone to add a record for. By default uses the one created by the DNS module if the module is found.
     default: ""
+  - name: forward_https
+    user_facing: true
+    validator: bool(required=False)
+    description: Should cloudformation forward to https port 443 instead of http port 80?
```
NitinAgg, line 100: So the user needs to set this to true if they are using the dns module?

rabbitfang (commented):
Small concern that might cause us issues

modules/cloudfront_distribution/cloudfront-distribution.yaml
juandiegopalomino: They don't _need_ to set it to true, they can set it to true if their lb has a valid ssl cert (either by using the one opta created by setting delegated true, or one they create themselves)

NitinAgg (approved):

Issue #876, opened 2022-05-11
Add features to k8s-service module


jocchionero:
#### What would you like to be added:
1. Would like dashes allowed in k8-service module, current error is:
```
ERROR: Invalid module name, can only contain letters and numbers!
If you need more help please reach out to the contributors in our slack channel at: https://slack.opta.dev
```

2. Would like to have custom health checks in same k8s-service module like in [standard k8s yaml](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
```
  - name: my-service
    type: k8s-service
...
    livenessProbe:
      exec:
        command:
        - cat
        - /tmp/healthy
...
```

#### Why is this needed:
1. This is a request from my team to stay consistent with current environments (pre opta migration)
2. A request from my team to allow this while we create actual health-check endpoints AND for 3rd part containers where we do not control the code to expose endpoints

### Extra info (e.g. existing slack convo link):
https://opta-group.slack.com/archives/C0240NBKUMT/p1652223456899139

juandiegopalomino:
release is out

PR #890, opened 2022-05-18
Add the ability to pass custom pod labels to k8s service

juandiegopalomino:
# Description
Like what we did for pod annotations

# Safety checklist

manually


juandiegopalomino:
https://github.com/run-x/opta/issues/876

RemyDeWolf (approved):

PR #921, opened 2022-07-07
Fix issue with public_uri when using var substitution with a list value

rabbitfang:
# Description
When a list var is used via variable substitution on the `k8s-service` `public_uri` input, terraform receives a nested list (e.g. `var=[foo, bar]` results in terraform getting `public_uri=[[foo, bar]]`).

# Safety checklist

YOUR_ANSWER


rabbitfang:
This change unfortunately does not actually fix the issue. The nested list is converted to a string before terraform even sees it.

